PR #11, opened 2020-08-19
Add example CloudFormation

katebee:

Add example templates that can create (almost) all the AWS resources necessary for running Janus in an AWS account.

Add links from the README to the example CloudFormation and provide guidance on how the templates can be used. 

Increase the number of CloudFormation templates in Open Source 🚀✨

Reduce the time investment for other organisations to adopt Janus.

Promote some good practices via example CloudFormation.

## Any additional notes?

Also fixing a small spelling error found in the tests. 

I have tested both the templates at commit `bbdd8f0` after commenting out the invalidating lines. CloudFormation validation passes and the stacks will successfully create:

Screenshot 2020-09-09 at 12 23 27


katebee:
I have rebased this PR and fixed the conflict that due to the recent update of the libraries and dependencies.

docs/cloudformation/federation.template.yaml
```diff
@@ -0,0 +1,60 @@
...
+    Properties:
+      Path: /federation/
+      MaxSessionDuration: 36000
+      AssumeRolePolicyDocument:
+        Statement:
+        - Effect: Allow
+          Principal:
+            AWS:
+            - !Ref JanusInstanceRoleArn
+            - !If [ HasDevUser, !Ref JanusDevUserArn, !Ref 'AWS::NoValue' ]
```
adamnfish, line 33: This is neat, but it looks like this condition (`HasDevUser) is not defined anywhere. The documentation tells me we'd need to add a `Conditions` section to the top level of the template that includes a `HasDevUser` entry. In practice, it may also be simpler to invert this Boolean so we can compare to the empty string to define the condition without also using `!Not`? No harm though, this is what `!Not` is for.

    Conditions:
      HasDevUser:
        !Not !Equals [ !Ref JanusDevUserArn, "" ]

(or inverted to avoid the `!Not` with a change to this `!If` statement if so)

This probably needs testing, I've not used conditions before so be good to see it working in AWS.

docs/cloudformation/federation.template.yaml
```diff
@@ -0,0 +1,60 @@
...
+# !! This template is provided as an example and is not intended for production use !!
+--!! This line invalidates the template and will need to be removed before CloudFormation !!--
+
+Parameters:
+  JanusInstanceRoleArn:
+    Description: ARN of Janus' IAM role
+    Type: String
+    AllowedPattern: "arn:aws:iam::[0-9]{12}:.+"
+    ConstraintDescription: Should be an AWS IAM role
+  JanusDevUserArn:
```
adamnfish, line 13: While I understand the need for this, it does punch a hole in the security of the application. I think I'd prefer to separately document how to do this for people that want to, rather than have it be explicitly a part of the template. I think this makes it too easy for people to give Janus access to an IAM user, which is somewhat at odds with the spirit of the application. In particular, it's unlikely that this is needed "be default". Partly because not everyone will use it, but also because even if they do, they are unlikely to use it for all stages or even all AWS accounts. As such, making this an exception rather than the norm might help guide users towards making better use of Janus.

Very happy to chat this one over with you. The local development issue is definitely something important to consider and I do agree that this approach is a good solution to that problem.

README.md
```diff
@@ -61,26 +61,42 @@ The configuration section (below) shows what is required.
...
+0. Modify the [example template](docs/cloudformation/example.template.yaml)
+to provide valid UserData for the launch configuration and remove the
+line that invalidates the template
 0. Use AWS Certificate Manager to create an HTTPS certificate for the domain
-0. Create CloudFormation stack using the supplied template
+0. Setup configuration as described below
+0. Create a CloudFormation stack using your modified template
 0. Point your DNS at Janus' Load Balancer
-0. Use the supplied federation template to integrate Janus with AWS accounts
+0. Use the [federation template](docs/cloudformation/federation.template.yaml) template to integrate Janus with AWS accounts
```
adamnfish, line 81: Should it be called `federation-example.template.yaml` to match the main template's name? No harm either way!

adamnfish (approved):
A few comments, but honestly nothing super important. Happy to chat about any of this, but equally happy if none of it feels worth addressing here.

Be good to chat about the local dev helper in the federation template, but no harm if you want to merge first and talk later.

docs/cloudformation/example.template.yaml
```diff
@@ -0,0 +1,287 @@
...
+  KeyName:
+    Description: Name of the SSH keypair to put on the instance
+    Type: AWS::EC2::KeyPair::KeyName
+  Stage:
+    Description: Environment name
+    Type: String
+    AllowedValues:
+      - PROD
+  AMI:
+    Description: Base AMI for Janus instances
```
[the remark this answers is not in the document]
adamnfish: It occurs to me that without the userdata, this may not be the case. The only thing is that the build.sbt already includes a bunch of code for producing an installable `deb` file for running Janus (which assumes a Debian derivative like Ubuntu).

docs/cloudformation/example.template.yaml
```diff
@@ -0,0 +1,287 @@
...
+        - AttributeName: j_username
+          AttributeType: S
+        - AttributeName: j_timestamp
+          AttributeType: N
+      KeySchema:
+        - AttributeName: j_account
+          KeyType: HASH
+        - AttributeName: j_timestamp
+          KeyType: RANGE
+      ProvisionedThroughput:
```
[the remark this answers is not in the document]
katebee: This is really good to know, and would apply to some other tooling too 👍

docs/cloudformation/federation.template.yaml
katebee: Very good points! ⭐  
I think it is best to remove the idea of a Dev user and wait to attempt it in the future if this is something that we want to support. I agree that it would actually be better as the exception, as end-to-end testing should be possible with access to only a couple of accounts.

docs/cloudformation/federation.template.yaml
katebee: Good catch, I must have missed out committing the Conditions 🤦 
I am removing the idea of the Dev User for now as I agree with your points around not making it the default approach.

PR #12, opened 2020-08-27
Update versions of Scala and library dependencies

adamnfish:

Update library dependencies, particularly the Play Googleauth dependency so that we take advantage of [an important security fix from the Google Auth library](https://snyk.io/vuln/SNYK-JAVA-COMGOOGLEOAUTHCLIENT-575276).

Up-to date software without security problems, with the application still working as expected.

## Any additional notes?

A separate PR will follow to update the way Play-Googleauth is used, for now the deprecated constructors are still being called. To prevent this causing compile errors, the PR also bumps the project to Scala 2.13.x so we get the `@nowarn` annotation. This change meant quite a few dependencies needed updating and it spiralled from there.

Of note, is that `scalatest` moved a few things around with v3.2 so the test changes are quite verbose. For the most part it is just renaming, but we've also had to introduce a few helpers to patch things that aren't working in scalatest on Scala 2.13. The `RightValues` and `HaveMatchers` utilities patch old behaviour back into the test suite to keep the test changes small.

## Deployment

I have tested this locally, and it looks to be fine. I'll deploy this to a real AWS environment as well to check it works fine in a production setting prior to merging. Let's do a code review first though!

### DIFF
--- a/app/conf/Config.scala
+++ b/app/conf/Config.scala
@@ -10,6 +10,7 @@ import models._
 import play.api.Configuration
 import play.api.http.HttpConfiguration
 
+import scala.annotation.nowarn
 import scala.util.Try
 
 
@@ -49,12 +50,16 @@ object Config {
     val clientSecret = requiredString(config, "auth.google.clientSecret")
     val domain = requiredString(config, "auth.domain")
     val redirectUrl = s"${requiredString(config, "host")}/oauthCallback"
+
+    @nowarn
+    val legacyAntiForgeryChecker = AntiForgeryChecker.borrowSettingsFromPlay(httpConfiguration)
+
     GoogleAuthConfig(
       clientId = clientId,
       clientSecret = clientSecret,
       redirectUrl = redirectUrl,
-      domain = domain,
-      antiForgeryChecker = AntiForgeryChecker.borrowSettingsFromPlay(httpConfiguration)
+      domains = List(domain),
+      antiForgeryChecker = legacyAntiForgeryChecker,
     )
   }
 
@@ -62,6 +67,7 @@ object Config {
     val twoFAUser = requiredString(config, "auth.google.2faUser")
     val serviceAccountCertPath = requiredString(config, "auth.google.serviceAccountCertPath")
 
+    @nowarn
     val credentials: GoogleCredential = {
       val jsonCertStream =
         Try(new FileInputStream(serviceAccountCertPath))
--- a/build.sbt
+++ b/build.sbt
@@ -13,21 +13,22 @@ ThisBuild / scmInfo := Some(ScmInfo(url("https://github.com/guardian/janus-app")
 ThisBuild / homepage := scmInfo.value.map(_.browseUrl)
 ThisBuild / developers := List(Developer(id = "guardian", name = "Guardian", email = null, url = url("https://github.com/guardian")))
 
-val awsSdkVersion = "1.11.663"
-val awscalaVersion = "0.8.1"
-val circeVersion = "0.12.3"
+val awsSdkVersion = "1.11.848"
+val awscalaVersion = "0.8.4"
+val circeVersion = "0.13.0"
 val commonDependencies = Seq(
   "org.typelevel" %% "cats-core" % "2.0.0",
-  "joda-time" % "joda-time" % "2.10.5",
+  "joda-time" % "joda-time" % "2.10.6",
   "org.joda" % "joda-convert" % "2.2.1",
   "com.github.seratch" %% "awscala-iam" % awscalaVersion,
   "com.github.seratch" %% "awscala-sts" % awscalaVersion,
   "com.github.seratch" %% "awscala-dynamodb" % awscalaVersion,
-  "org.scalatest" %% "scalatest" % "3.0.5" % Test,
-  "org.scalacheck" %% "scalacheck" % "1.14.0" % Test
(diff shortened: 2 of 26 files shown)


app/conf/Config.scala
```diff
@@ -49,19 +50,24 @@ object Config {
     val clientSecret = requiredString(config, "auth.google.clientSecret")
     val domain = requiredString(config, "auth.domain")
     val redirectUrl = s"${requiredString(config, "host")}/oauthCallback"
+
+    @nowarn
```
adamnfish, line 54: These `@nowarn` calls are required because we have enabled `-Xfatal-warnings` for this build, in line with Scala Best Practices. Deprecation is a warning, but with `-Xfatal-warnings` enabled this is a compile error. Luckily, Scala 2.13 introduces `@nowarn` so we can temporarily ignore these warnings during this multi-step upgrade process.

build.sbt
```diff
@@ -57,7 +58,7 @@ lazy val root = (project in file("."))
     libraryDependencies ++= commonDependencies ++ Seq(
       ws,
       filters,
-      "com.gu" %% "play-googleauth" % "0.7.4",
+      "com.gu.play-googleauth" %% "play-v27" % "1.0.7",
```
adamnfish, line 61: This is the key update. We'll switch over to use the non-deprecated use of this version in a subsequent PR so that the details aren't hidden by upgrade boiler-plate.

configTools/src/test/scala/com/gu/janus/testutils/HaveMatchers.scala
```diff
@@ -0,0 +1,18 @@
+package com.gu.janus.testutils
+
+import org.scalatest.matchers.HavePropertyMatcher
+import org.scalatest.matchers.should.Matchers
+
+
+trait HaveMatchers extends Matchers {
```
adamnfish, line 7: This helper restores the super-useful `have` matcher for Scala 2.13. Symbols are deprecated so you can no longer write this the way scalatest expects.

Old way:
```scala
foo should have(
  'fieldName (value),
  'fieldName2 (value2),
)
```

New way
```scala
foo should have(
  "fieldName" as value,
  "fieldName2" as value2,
)
```

configTools/src/test/scala/com/gu/janus/testutils/RightValues.scala
```diff
@@ -0,0 +1,24 @@
+package com.gu.janus.testutils
+
+import org.scalactic.source.Position
+import org.scalatest.exceptions.{StackDepthException, TestFailedException}
+
+trait RightValues {
```
adamnfish, line 6: This restore `EitherValues` behaviour that works with fully Right-biased Eithers. Scala 2.13 deprecates using the Right projection, which is what EitherValues still requires us to use. `RightValues` adds `.value` to Either types directly.

Old way
```scala
val foo: Either[L, R] = ???
foo.right.value shouldEqual expected
```

New way
```scala
val foo: Either[L, R] = ???
foo.value shouldEqual expected
```

Handling Left values can still be done using `EitherValues`, because the LeftProjection still works fine.

project/plugins.sbt
```diff
@@ -5,7 +5,7 @@ addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.9")
 addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.3.6")
 
 // The Play plugin
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.7.3")
+addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.7.5")
```
adamnfish, line 8: Also update Play while we're here. A 2.8.x update can follow separately.

project/plugins.sbt
```diff
@@ -14,4 +14,6 @@ libraryDependencies += "org.vafer" % "jdeb" % "1.3" artifacts Artifact("jdeb", "
 // These are for releasing to Sonatype
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.9.2")
-addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.10")
\ No newline at end of file
+addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.10")
+
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
adamnfish, line 19: I used this during testing, figure I may as well leave it here. I don't feel strongly about this though, happy to remove it?

adamnfish (commented):
Here are some guidelines for review, these are points of interest among the noise of mechanically renaming imports and classes.

katebee (approved):
👍 ⭐

project/plugins.sbt
philmcmahon: I would lean towards this going in your local ~/.sbt/ plugins settings rather than at a project level, not a big deal though

akash1810 (approved):
Great stuff! 👍🏽

project/plugins.sbt
adamnfish: Thanks @philmcmahon, you're absolutely right. That would be a much better approach, I'll remove this today.

adamnfish:
I have rebased this PR to include the updates for @katebee's recently merged change to include the `Statements` in the published configTools (https://github.com/guardian/janus-app/pull/13).

PR #13, opened 2020-08-27
Move statements into configTools

katebee:

Move Statements out of example and into configTools, and bump the version of configTools depended upon.
The Statements are a useful shared functionality that can be used for creating bespoke policies.

This also adds the required `permissionsRepo` to the JanusData used in the example app.

We can release a new version of configTools


adamnfish (approved):
Looks great, thanks for doing this.

PR #15, opened 2020-09-04
Add method to verify the janusData can be found and loaded

katebee:

Add support for verification that the JanusData can be found and loaded.

This can be ran as part of a build script and will provide helpful information if an error is encountered and prevent the build from continuing.

Janus will only be deployed if the janusData is present and valid. 

## Images

Example of error printed before a non-zero exit:

Screenshot 2020-09-04 at 11 50 58

### DIFF
--- a/configTools/src/main/scala/com/gu/janus/VerifyJanusData.scala
+++ b/configTools/src/main/scala/com/gu/janus/VerifyJanusData.scala
@@ -12,9 +12,9 @@ object VerifyJanusData {
   def main(args: Array[String]): Unit = {
     args.headOption.fold {
       Console.err.println(
-        s"""${RED}Error: Missing argument <output file>
+        s"""${RED}Error: Missing argument <input file>
            |
-           |Usage: example <output file>$RESET""".stripMargin)
+           |Usage: run <input file>$RESET""".stripMargin)
       System.exit(1)
     } { fileName =>
       try {
@@ -35,7 +35,7 @@ object VerifyJanusData {
           System.exit(1)
         case NonFatal(e) =>
           Console.err.println(
-            s"""${RED}Error loading JanusData: ${e.getMessage}>$RESET""".stripMargin)
+            s"""${RED}Error loading JanusData: ${e.getMessage}$RESET""".stripMargin)
           System.exit(1)
       }
     }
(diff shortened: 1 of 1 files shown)


configTools/src/main/scala/com/gu/janus/VerifyJanusData.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import scala.Console.{GREEN, RED, RESET}
+import scala.util.control.NonFatal
+
+
+object VerifyJanusData {
+  def main(args: Array[String]): Unit = {
+    args.headOption.fold {
+      Console.err.println(
+        s"""${RED}Error: Missing argument <output file>
```
adamnfish, line 15: I think `output file` should instead refer to the janus data configuration input file?

configTools/src/main/scala/com/gu/janus/VerifyJanusData.scala
```diff
@@ -0,0 +1,43 @@
...
+           |
+           |Usage: example <output file>$RESET""".stripMargin)
```
adamnfish, line 17: Same here

configTools/src/main/scala/com/gu/janus/VerifyJanusData.scala
```diff
@@ -0,0 +1,43 @@
...
+          System.exit(1)
+        }
+      } catch {
+        case e: JanusConfigurationException =>
+          Console.err.println(
+            s"""$RED${e.getMessage}$RESET""".stripMargin)
+          System.exit(1)
+        case NonFatal(e) =>
+          Console.err.println(
+            s"""${RED}Error loading JanusData: ${e.getMessage}>$RESET""".stripMargin)
```
adamnfish, line 38: Is the `>` symbol meant to be there?

configTools/src/main/scala/com/gu/janus/VerifyJanusData.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import com.gu.janus.JanusConfig.JanusConfigurationException
+
+import scala.Console.{GREEN, RED, RESET}
+import scala.util.control.NonFatal
+
+
+object VerifyJanusData {
+  def main(args: Array[String]): Unit = {
+    args.headOption.fold {
```
adamnfish, line 13: It occurs to me now, that a `match` might be easier to read for people that aren't used to `Option.fold` here. No biggie, I like `Option.fold` myself so this is totally fine as-is, up to you.

adamnfish (approved):
Great, thank you.

PR #17, opened 2020-09-25
Add more details on how to publish configTools

katebee:

Expand the README how to update the version number of configTools in the main branch after publishing a release.

Ensure someone following the readme switches the the correct project before attempting to release configTools.
Encourage pushing git tags along with branch to the remote so that releases of configTools are tagged.


adamnfish (approved):
Great addition, thanks for fixing these instructions.

PR #18, opened 2020-10-22
Add first examples of bespoke policies

katebee:

To add a few examples of bespoke polices along with comments about their purpose.

This also adds another employee to our example company to show how bespoke polices can be added to a user.

Examples should demonstrate how to create policies that apply to specific resources in fixed accounts, along with some reasonable use cases for them.

## Any additional notes?

We are beginning to run out of characters from Sir ACD 🤷 

It remains to be seen if employing both Sherlock and Mycroft is a good idea, but they will both have to just...

mycroft

😎 😅 😎 😅 😎


adamnfish (approved):
Great, thank you!

PR #19, opened 2021-08-18
[Do NOT Merge] Add janus data to riff raff artifacts

jfsoul:
This is a draft idea for publishing Janus config to an alternate S3 location in order to facilitate auditing by another process.


jfsoul:
Closing in favour of a different approach for now.

PR #21, opened 2021-08-23
Add console logging

jacobwinch:

* This change allows us to send logs to `stdout` as well as a log file. This allows us to collect logs via [Fluent Bit's systemd plugin](https://docs.fluentbit.io/manual/pipeline/inputs/systemd), which is an approach we're exploring for shipping logs at the OS level (rather than via changes to the application code).

* Application logs should be available via `journalctl -u janus.service`
* When combined with the [config added here](https://github.com/guardian/janus/pull/2774), we should be able to ship logs to Central ELK

### DIFF
--- a/conf/logback.xml
+++ b/conf/logback.xml
@@ -15,6 +15,15 @@
         </encoder>
     </appender>
 
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%date [%level] from %logger in %thread - %message%n%xException</pattern>
+        </encoder>
+    </appender>
+
+    <appender name="ASYNCSTDOUT" class="ch.qos.logback.classic.AsyncAppender">
+        <appender-ref ref="STDOUT" />
+    </appender>
 
     <logger name="play" level="INFO" />
     <logger name="application" level="DEBUG" />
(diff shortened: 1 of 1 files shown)


conf/logback.xml
```diff
@@ -15,6 +15,15 @@
         </encoder>
     </appender>
 
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%date [%level] from %logger in %thread - %message%n%xException</pattern>
```
jfsoul, line 20: I don't know what the right format is off the top of my head, but I assume this will work for our purposes

jfsoul (approved):
Presumably this `logback.xml` also gets used locally - in which case my only question is if this change is desirable to apply locally? I assume the dev-ex is fine, but thought I'd ask!

jacobwinch:
> in which case my only question is if this change is desirable to apply locally? I assume the dev-ex is fine, but thought I'd ask!

Personally I find it reasonably useful to have logging in the console (I'd usually end up `tail`ing a log file anyway otherwise), but it's pretty trivial to redirect `stdout` (to another file or `/dev/null`) for those who might disagree! It does look like there are some more [options available](https://www.playframework.com/documentation/2.8.x/SettingsLogger#Custom-configuration) if we wanted local vs prod specific configurations, but personally I'm not sure whether it's worth the additional complexity.

conf/logback.xml
jacobwinch: In the longer-term I think we will be better off using structured/JSON log lines, but this should be OK for the proof of concept!

PR #23, opened 2021-12-10
Upgrade SBT to 1.5.6 (edit to 1.5.7 and Play from 2.7 to 2.8)

nicl:
Immediate motivation was to use the dependencyTree command to check for Log4j usage (re recent security vulnerability). But this is generally useful in any case.

See also: https://github.com/guardian/janus/pull/2904.

(Not really tested beyond CI - any suggestions on best way to test this appreciated.)

**Note: the Play upgrade would ideally be in a separate PR, but is forced by https://github.com/guardian/ophan/commit/bf8576e5c3a3c813de8b64271d3db260c7befbdf.**


jacobwinch (approved):

adamnfish (commented):
This PR makes lots of changes! Any chance you could re-open it with just an sbt bump, or at least rename it to a more general "update sbt and play, linting"? The linting changes are a distraction but not urgent, but I suspect bumping the Play version is something we'd want to pay attention to and be able to roll back independently. If we want to do all this together it'd be good for people to be able to easily find this PR if they're looking for why some part of the Play configuration is now broken.

PR #137, opened 2022-11-14
Bump dependencies on example project

kenoir:

The example project here was being flagged by vulnerability monitoring, also upon inspection it did not appear to compile / test successfully.

This change bumps some dependencies in the example project to provide better defaults for someone using the project and removes some code that was preventing the project compiling.

### DIFF
--- a/example/src/main/scala/com/example/Access.scala
+++ b/example/src/main/scala/com/example/Access.scala
@@ -16,7 +16,7 @@ object Access {
     "sherlock.holmes" -> (Root.dev ++ securityAccess),
     "john.watson" -> (Production.dev ++ Staging.dev ++ Root.billing),
     "irene.adler" -> (Production.dev ++ Staging.dev ++ DataLake.dev),
-    "mycroft.holmes" -> (productionAccessLogs ++ securityCloudtrailLogs ++ serviceControlPolicyAccess)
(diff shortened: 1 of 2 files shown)


example/src/main/scala/com/example/Access.scala
```diff
@@ -16,7 +16,7 @@ object Access {
     "sherlock.holmes" -> (Root.dev ++ securityAccess),
     "john.watson" -> (Production.dev ++ Staging.dev ++ Root.billing),
     "irene.adler" -> (Production.dev ++ Staging.dev ++ DataLake.dev),
-    "mycroft.holmes" -> (productionAccessLogs ++ securityCloudtrailLogs ++ serviceControlPolicyAccess)
+    "mycroft.holmes" -> (productionAccessLogs ++ securityCloudtrailLogs)
```
akash1810, line 19: Why is `serviceControlPolicyAccess` dropped?

akash1810 (approved):
LGTM - one minor, non-blocking, comment.

example/src/main/scala/com/example/Access.scala
kenoir: It doesn't exist, so was causing the project to fail to compile. Looking again I believe that this should actually refer to `BespokePolicies.serviceControlPolicyManagement` so have added that back in to include it in the example.

akash1810 (approved):

PR #187, opened 2023-01-25
Scalafmt

nicl:
We are migrating all DevX Scala tooling to Scalafmt. See e.g. https://github.com/guardian/riff-raff/pull/940 and also this Trello card: https://trello.com/c/eIgRRUgB/1634-introduce-scalafmt-to-our-scala-repos.

Note, the actual CI check here is done by a script in the /guardian/janus repo (as Teamcity combines both repos for the build). So will raise a follow-up PR for that.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,4 @@
+version = 3.6.1
+
+runner.dialect = scala213
+maxColumn = 80
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,8 @@ atlassian-ide-plugin.xml
 com_crashlytics_export_strings.xml
 crashlytics.properties
 crashlytics-build.properties
+
+# VS Code
+.bsp/
+.metals/
+.vscode/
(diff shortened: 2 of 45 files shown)


.scalafmt.conf
```diff
@@ -0,0 +1,4 @@
+version = 3.6.1
+
+runner.dialect = scala213
+maxColumn = 80
```
akash1810, line 4: 80 chars wide, on 2023 screens is very small...

akash1810 (approved):

.scalafmt.conf
nicl: Agreed it is small! But there are some good reasons for it:

* it is how we format our TS/JS code (https://github.com/guardian/csnx/blob/main/libs/%40guardian/prettier/index.js#L6) and consistent with Prettier [recommendations](https://prettier.io/docs/en/options.html#print-width)
* human eyes find this width easy to scan (think of paperback books which are typically sized for easy reading)
* easier to have two editors/windows side-by-side
* easier to see entire lines without wrapping in Github code reviews
* (finally, and less important, we've already applied this to all of our DevX scala repos now and would need to reformat them again to change it)

PR #314, opened 2023-06-16
Migrate teamcity pipeline to github actions

Mark-McCracken:
## What is this change for?
Calls the internal guardian build process, to validate this steps, which was previously based on a different internal build system

### DIFF
--- a/.github/workflows/trigger-a-build-elsewhere.yml
+++ b/.github/workflows/trigger-a-build-elsewhere.yml
@@ -22,7 +22,7 @@ jobs:
             workflow_id: 'build.yml',
             ref: 'mmc/migrate-build-from-teamcity-to-gha',
             inputs: {
-              "janus-app-branch": ${{ env.GITHUB_REF_NAME }}
+              "janus-app-branch": ${{ github.ref_name }}
             }
           })
     
--- a/.github/workflows/trigger-a-build-elsewhere.yml
+++ b/.github/workflows/trigger-a-build-elsewhere.yml
@@ -1,0 +1,29 @@
+name: Trigger Janus build
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Trigger workflow
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.PAT_actions_janus }}
+        script: |
+          await github.rest.actions.createWorkflowDispatch({
+            owner: 'guardian',
+            repo: 'janus',
+            workflow_id: 'build.yml',
+            ref: 'mmc/migrate-build-from-teamcity-to-gha'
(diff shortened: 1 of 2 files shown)


.github/workflows/trigger-a-build-elsewhere.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Trigger workflow
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.PAT_ACTIONS_JANUS }}
```
nicl, line 17: Worth commenting this and perhaps linking to docs on how to regenerate when it expires?

nicl (approved):

.github/workflows/trigger-a-build-elsewhere.yml
```diff
@@ -0,0 +1,32 @@
...
+            repo: 'janus',
+            workflow_id: 'build.yml',
+            ref: 'mmc/migrate-build-from-teamcity-to-gha',
+            inputs: {
+              "janus-app-branch": "${{ github.ref }}"
+            }
+          })
+    
+    - name: Please check janus repo for build status.
+      run: echo "Please check janus repo for status"
```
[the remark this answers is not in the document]
Mark-McCracken: New design makes this idea less relevant

.github/workflows/trigger-a-build-elsewhere.yml
Mark-McCracken: documented this in the script

.github/workflows/trigger-private-janus-build.yml
```diff
@@ -0,0 +1,90 @@
...
+            owner: 'guardian',
+            repo: 'janus',
+            workflow_id: 'build.yml',
+            event: 'workflow_dispatch',
+            created: '>=${{ env.START_TIME }}'
+          });
+          console.log(result);
+          return result["data"]["workflow_runs"][0].id
+
+    - run: sleep 240 # wait 4 minutes for job to run
```
akash1810, line 65: How was this duration chosen?

Mark-McCracken: Oh wow, the post setup action for caching sbt dependencies has massively increased. I selected the 4 minutes based on previous runs in teamcity being about 2 and a half minutes, and the github actions runs being about a minute longer due to caching the dependencies, but this run seems like it took 4 mins to cache the dependencies, I wonder if we've added a much larger set of packages?

On the plus side, this is why I broke the workflow into 2 jobs, so that if the time is exceeded by the build, we just need to re-run the failed job that checks for deployment

Mark-McCracken: @akash1810 I've just re-run the failed job, and it completed successfully.
I think we should wait and see if subsequent builds are also much longer, and if so, bump this wait time

PR #347, opened 2023-08-18
chore(deps): Rollback update to `logstash-logback-encoder`

akash1810:
Fix an issue where logs are not appearing in Central ELK.

Running Janus locally, we're getting the following error:

```log
Exception in thread "AsyncAppender-Worker-ASYNCSTDOUT" java.lang.NoSuchMethodError: 'java.time.Instant ch.qos.logback.classic.spi.ILoggingEvent.getInstant()'
```

Rolling back fixes this.

Logs will start appearing in Central ELK again.


akash1810:
Confirming that applications logs are landing in Central ELK again now - https://logs.gutools.co.uk/s/devx/goto/b8b1e1c0-3de3-11ee-a76a-8328e483668b.

PR #365, opened 2023-10-03
ci: Tighten the build

akash1810:
This change removes the `janus-app-branch` input to the Workflow being triggered in `guardian/janus`, removing an edge case where feature branches were being deployed to PROD.

The build is now:
  1. Compile the Scala app, and run the tests
  2. If compilation is successful, and we're on the `main` branch, trigger the Workflow in `guardian/janus`
  3. The Workflow in `guardian/janus` will clone always clone the main branch of this repository

With these changes we lose the ability to deploy feature branches (as their build will not appear in Riff-Raff), however looking at the deployment history, this isn't a thing we've ever done.

These changes removes an edge case from the build where feature branches in this repository were being deployed to PROD.

## Images

Here's a diagram of the build steps:

image

---

See also https://github.com/guardian/janus/pull/3769.

### DIFF
--- a/.github/workflows/trigger-private-janus-build.yml
+++ b/.github/workflows/trigger-private-janus-build.yml
@@ -1,24 +1,53 @@
-# This triggers a build in a separate github action
-
-## See documentation here: https://docs.google.com/document/d/1Xz-gzptICIza0oio0Kza3LWg2a6n_015o0MPu0LGfuU/edit#heading=h.k45jb8zdijoi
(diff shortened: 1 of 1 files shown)


AshCorr (approved):
Looks good!

Curious on why we need to run tests now, is this because previously the private janus repo would have handled testing?

akash1810:
> Curious on why we need to run tests now, is this because previously the private janus repo would have handled testing?

Yep.

PR #385, opened 2023-12-18
Fix: downgrade play to make sbt work again

tjsilver:
Downgrades the Play plugin from 2.9.0 (it was upgraded in #372) to 2.8.19, to fix issues:
1. SBT was no longer working locally
2. The `Trigger Private Janus build` action did not run successfully

SBT can be run without error locally (and subsequently `snyk test`)

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -14,7 +14,7 @@ ThisBuild / libraryDependencySchemes ++= Seq(
 )
 
 // The Play plugin
-addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.9.0")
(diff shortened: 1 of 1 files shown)


akash1810:
Worth adding a stack trace, describing what happens on Play 2.9.0?

Can we make any changes to the build to catch this and similar errors, before they are merged?

tjsilver:
> Worth adding a stack trace, describing what happens on Play 2.9.0?
> 
> Can we make any changes to the build to catch this and similar errors, before they are merged?

Good idea - not sure what this involves - can we do that in a follow-up PR @akash1810 ?

jacobwinch (approved):
Let's merge this in and add more details as a follow-up - IIUC Janus cannot be built/deployed at the moment?

PR #393, opened 2024-01-09
chore: Upgrade to Java 11

tjsilver:
This is a re-attempt to upgrade Java to v11. This was attempted in #379 and reverted in #381.

Requires: https://github.com/guardian/janus/pull/3947 which should be merged first, as Java 11 is backwards compatible with Java 8.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -114,7 +114,6 @@ lazy val root = (project in file("."))
     Compile / packageDoc / publishArtifact := false,
     Debian / topLevelDirectory := Some(normalizedName.value),
     Debian / serverLoading := Some(Systemd),
-    debianPackageDependencies := Seq("java8-runtime-headless"),
     Debian / maintainer := "Developer Experience <dig.dev.tooling@theguardian.com>",
     Debian / packageSummary := "Janus webapp",
     Debian / packageDescription := "Janus: Google-based federated AWS login"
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -121,7 +114,6 @@ lazy val root = (project in file("."))
     Compile / packageDoc / publishArtifact := false,
     Debian / topLevelDirectory := Some(normalizedName.value),
     Debian / serverLoading := Some(Systemd),
-    debianPackageDependencies := Seq("java8-runtime-headless"),
```
tjsilver, line 124: This was removed in the previous attempt, but I'm not sure why it's no longer required or doesn't need upgrading.

build.sbt
```diff
@@ -88,15 +88,8 @@ lazy val root = (project in file("."))
     Universal / javaOptions ++= Seq(
       "-Dconfig.file=/etc/gu/janus.conf", // for PROD, overridden by local sbt file
       "-Dpidfile.path=/dev/null",
-      "-J-XX:MaxRAMFraction=2",
```
tjsilver, line 91: The error in the last attempt was thought to be caused by `"-J-XX:+PrintGCDateStamps"`, and there were warnings caused by the RAM fraction options`-J-XX:MaxRAMFraction=2` and `-J-XX:InitialRAMFraction=2`. The RAM fractions have been replaced with initial heap size(`Xms`) and max heap size (`Xmx`). After discussion with @adamnfish and @jacobwinch all the advanced options (`-XX`) were deemed unnecessary and thus removed.

build.sbt
tjsilver: @adamnfish has suggested that it "predates our base images ... it would have been used on ubuntu images that installed everything at start time"

adamnfish (dismissed):
As mentioned below, maybe we tweak the RAM amount to match what was there for now?

jacobwinch (approved):

PR #398, opened 2024-01-17
Chore: bump config tools after releasing to Maven

tjsilver:

This bumps the version of `janus-config-tools` to v0.0.6. This reflects the version of the latest Maven release.  

The published version has its dependency updates and the versions are in line between Maven and this repo.

Co-authored-by: @adamnfish


adamnfish (approved):
We've seen this artifact in maven 👍

PR #400, opened 2024-01-18
ci: Use first-party GitHub Action 

akash1810:
Following on from #396, this change uses a first-party[^1] GitHub Action to exchange a GitHub App private key for a token.

Similar to #396, this has been tested by temporarily removing the [if condition](https://github.com/guardian/janus-app/blob/6118e23bf855f1bf8579e55209093a6c2f39356e/.github/workflows/trigger-private-janus-build.yml#L49):

image

[^1]: See also https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions.


.github/workflows/trigger-private-janus-build.yml
```diff
@@ -58,18 +58,20 @@ jobs:
...
       with:
         # These values are for the GitHub App guardian-janus-ci
-        # See https://github.com/organizations/guardian/settings/apps/guardian-janus-ci (only accesssible by GitHub owners)
-        app_id: ${{ secrets.GH_APP_ID }} # The ID of the GitHub App
-        private_key: ${{ secrets.GH_APP_PRIVATE_KEY }} # The private key of the GitHub App
-        
+        # See https://github.com/organizations/guardian/settings/apps/guardian-janus-ci (only accessible by GitHub owners)
+        app-id: ${{ secrets.GH_APP_ID }}
+        private-key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+        owner: guardian
```
adamnfish, line 68: Does it work to use an org as the owner? It isn't clear to me from the docs that that is the case. Most likely you've already tested this works!

akash1810: This was the only way I could [get this to work](https://github.com/guardian/janus-app/actions/runs/7575229176/job/20631402703). My understanding is the resulting token is scoped exactly to the `guardian/janus` repository. If one tried to use it against a different repository, a 403 would be returned.

image

adamnfish (approved):
Looks great, and this is a very sensible change. One Q about the owner below, but you'll have read the docs in more detail / tested it, in which case LGTM!

.github/workflows/trigger-private-janus-build.yml
adamnfish: Yeah nice one, I can see the repositories key looks great. Thanks!

PR #402, opened 2024-01-31
feat: add sbt dependency submission action

tjsilver:

Adds a workflow that updates the dependency graph with SBT dependencies, using the third party (from Scalacenter) [SBT Dependency Submission workflow](https://github.com/marketplace/actions/sbt-dependency-submission).

This wil allow Dependabot to report on vulnerabilites in Scala dependencies. 

This has been tested on push to this branch and [created a snapshot of the dependencies](https://api.github.com/repos/guardian/janus-app/dependency-graph/snapshots/10509078). (See them under `snapshots > manifests > com.gu:janus... > resolved`).

When it is run on `main` after this PR is merged, the Scala dependencies should appear in [the dependency graph in this repo](https://github.com/guardian/janus-app/network/dependencies).

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -1,0 +1,12 @@
+name: Update Dependency Graph
+on:
+  push:
+    branches:
+      - main # default branch of the project
+jobs:
+  dependency-graph:
+    name: Update Dependency Graph
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: scalacenter/sbt-dependency-submission@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-graph.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  push:
+    branches:
+      - main # default branch of the project
+jobs:
+  dependency-graph:
+    name: Update Dependency Graph
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
akash1810, line 11: Should we use [v4 of `actions/checkout`](https://github.com/actions/checkout/blob/main/CHANGELOG.md)?

akash1810 (approved):

.github/workflows/dependency-graph.yml
tjsilver: Good spot, this was the workflow recommended on the [action's page](https://github.com/marketplace/actions/sbt-dependency-submission), so will try in a follow-up PR.

PR #407, opened 2024-02-07
fix: use later aws sdk version in configTools

tjsilver:

Makes the configTools project use the same version of `aws-sdk` as the main app. 

This is an attempt to mitigate a high severity vulnerability discovered by Dependabot, whereby configTools' reliance on [aws-scala](https://github.com/seratch/AWScala) which has been deprecated, and which in turn relies on an old version of `aws-java-sdk-s3`. Once this is merged, if successful we should see the issue disappear.

I have tested this locally and the dependency tree for configTools shows that the vulnerable version has been replaced a safe version:

Before: 
image

After: 
image


tjsilver:
Closing for now to see if Dependabot raises a PR

tjsilver:
Re-opening as Dependabot didn't raise a PR to fix this.

PR #409, opened 2024-02-08
fix: update sbt dependency submission workflow

tjsilver:

Removes test configs from the SBT dependency submission action.

We are not concerned with vulnerabilities that are not in production.

I tested this by running the action against this branch and checking the resulting snapshots of the graph. The final snapshot did not contain the htmlunit or jetty dependencies, so they should disappear from the vulnerability list in dependency graph once this is merged.


adamnfish (commented):
I think if we're doing this, we should probably lean into it and remove all the non-production configs.

Some others listed in the docs are "scala-tool" for compiler dependencies, "scala-doc-tool" for teams using scaladoc to generate documentation.

I imagine there's a config for the integration test scope, which is referred to as "it" in sbt. I can't see this mentioned anywhere in the plugin's source code, because this is an sbt thing (and to some extent a Scala convention). If we can think of anyone that's using this, maybe we'd be able to test it on their project, or we could test it ourselves by adding some dummy integration tests here?

This isn't saying we definitiely do this, let's chat about it?

.github/workflows/dependency-graph.yml
```diff
@@ -2,13 +2,16 @@ name: Update Dependency Graph
...
+      # - main # default branch of the project
 jobs:
   dependency-graph:
     name: Update Dependency Graph
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: scalacenter/sbt-dependency-submission@v2
+        with:
+          configs-ignore: test test-internal compile compile-internal scala-tool scala-doc-tool
```
tjsilver, line 15: Do we want to ignore `compile`/`compile-internal` config? It seems to get replaced by `runtime` config when ignored. @adamnfish

tjsilver: Actually, after discussion with @NovemberTang, we have decided to leave the workflow at default, and exclude any dev dependencies at the repocop stage, so that we can change the level of monitoring centrally (i.e. if we did decide to alert on dev dependencies in the future).

PR #416, opened 2024-03-19
Follows the instructions to enable the scala lib release workflow

adamnfish:
Removes some of the release-specific settings that are replaced by the workflow, and adds the required configuration for the workflow.

It should be easy to publish the config tools library.

Reducing friction on publishing this library will make it easier for us to make improvements.

## Any additional notes?

The PR that grants access has been merged, so we're free to do this when we have some time to check the release process. The fact that the library is a subfolder is unusual, so we'll pay special attention to that.


tjsilver (commented):
The CI status checks ruleset needs to be tweaked as scala steward PRs are not mergable currently - see https://github.com/guardian/janus-app/pull/415 - ticket created in Trello also

.github/workflows/release.yml
```diff
@@ -0,0 +1,13 @@
+name: Release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  release:
+    uses: guardian/gha-scala-library-release-workflow/.github/workflows/reusable-release.yml@use-java-21-lts
```
rtyley, line 8: Whoops, my fault - should be:
```suggestion
    uses: guardian/gha-scala-library-release-workflow/.github/workflows/reusable-release.yml@main
```

tjsilver (approved):

rtyley:
Looks like `janus-config-tools` has been successfully released with [`1.0.0`](https://github.com/guardian/janus-app/releases/tag/v1.0.0) - great work! 👍 

image

PR #429, opened 2024-04-16
docs: update README to reflect new release process

tjsilver:

Updates the README to reflect the new process for publishing Janus ConfigTools. We now publish using a GitHub Action rather than from SBT.


adamnfish (approved):
Thank you!

PR #439, opened 2024-05-17
add auto-logout toggle

twrichards:
When working with multiple AWS accounts throughout the day, it would be useful to have an (optional) auto-logout feature when clicking through to the console of the next/desired AWS account.

Turns out there's already an issue for this from 2017: https://github.com/guardian/janus/issues/868

## Changes

- allow `signin.aws.amazon.com` and `aws.amazon.com` as `connect-src`s in the CSP in anticipation of calling it client-side to perform auto-logout
- add an 'Auto-logout' toggle (persisted to local storage) and if active silently hit `https://signin.aws.amazon.com/logout` for AWS Console links before redirecting to the AWS console for desired/clicked account (gives up after 1.5s and redirects regardless)

#### Toggle
image

#### Example console output (from local)
image

#### Local Storage value (from local)
image

### DIFF
--- a/public/javascripts/janus.js
+++ b/public/javascripts/janus.js
@@ -279,4 +279,30 @@ jQuery(function($){
         }
     });
 
+    // auto-logout (preference persisted with local storage)
+    $("#auto_logout_switch").each(function(_, autoLogoutSwitchElement){
+        const LOCAL_STORAGE_KEY__AUTO_LOGOUT = "autoLogout"
+        autoLogoutSwitchElement.checked = localStorage.getItem(LOCAL_STORAGE_KEY__AUTO_LOGOUT) === "true";
+        autoLogoutSwitchElement.onchange = (event) => {
+            autoLogoutSwitchElement.checked = event.target.checked;
+            localStorage.setItem(LOCAL_STORAGE_KEY__AUTO_LOGOUT, event.target.checked.toString());
+        };
+
+        $("a[href*='/console?permissionId=']").each(function(_, el){
+            el.onclick = (clickEvent) => {
+                if(autoLogoutSwitchElement.checked) {
+                    clickEvent.preventDefault();
+                    const targetHref = el.href;
+                    console.log("Silently attempting logout before navigating to", targetHref)
+                    fetch("https://signin.aws.amazon.com/logout", {
+                        mode: "no-cors", // we avoid CORS issues here and really only care if the request succeeds,
+                        credentials: "include", // we need AWS cookies to be sent in this log out call
+                    }).finally(() => {
+                        location.href = targetHref;
+                    });
+                }
+            }
+        });
+    });
+
 });
--- a/public/javascripts/janus.js
+++ b/public/javascripts/janus.js
@@ -295,8 +295,12 @@ jQuery(function($){
                     const targetHref = el.href;
                     console.log("Silently attempting logout before navigating to", targetHref)
                     fetch("https://signin.aws.amazon.com/logout", {
-                        mode: "no-cors", // we avoid CORS issues here and really only care if the request succeeds,
-                        credentials: "include", // we need AWS cookies to be sent in this log out call
+                        // we avoid CORS issues here and really only care if the request succeeds
+                        mode: "no-cors",
+                        // we need AWS cookies to be sent in this log out call
+                        credentials: "include",
+                        // give up after short delay to ensure user intent is followed promptly of log out is slow
+                        signal: AbortSignal.timeout(1500),
                     }).finally(() => {
                         location.href = targetHref;
                     });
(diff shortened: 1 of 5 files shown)


public/javascripts/janus.js
```diff
@@ -279,4 +279,33 @@ jQuery(function($){
...
+    autoLogoutSwitch.prop('checked', shouldAutoLogout)
+    autoLogoutSwitch.on("change", () => {
+        shouldAutoLogout = event.target.checked;
+        localStorage.setItem(LOCAL_STORAGE_KEY__AUTO_LOGOUT, shouldAutoLogout.toString());
+    });
+
+    // auto-logout for console buttons
+    $("a[href*='/console?permissionId=']").each(function(_, el){
+        const targetHref = el.href;
+        el.removeAttribute("href");
```
adamnfish, line 295: I thought removing the `href` attribute disabled the link, which would mean it isn't focusable? If so, is it better to use preventDefault in the handler and leave the href attribute present? This also means things like dragging the link to another tab instead of clicking it would still work (with the caveat that this auto-logout would not work that way*).

* I think these kind of UX-problem edge cases are a necessary evil of this approach though!

twrichards: excellent point, thanks @adamnfish - will do that now

public/javascripts/janus.js
```diff
@@ -279,4 +279,33 @@ jQuery(function($){
         }
     });
 
+    // auto-logout switch (persisted with local storage)
+    const LOCAL_STORAGE_KEY__AUTO_LOGOUT = "autoLogout"
```
adamnfish, line 283: I think this will break on pages other than those that include the logout switch.

The approach the other "modules" take is to put the logic for each piece of functionality behind a check that this module is included on the current page, which has two benefits.

Firstly, this means that each piece of functionality works wherever it is included, without having to explicitly list the routes where it would apply. For example, the account picking functionality is included on the homepage, but also appears on the support and admin pages for people on support and people with admin access respectively. Rather than duplicate the logic for deciding whether the account picking logic should apply, this JS file is included everywhere and the setup for each module happens when its elements are found.
In this case we'd want to have a think about whether all of these account-picking pages should include the auto-logout behaviour, and if so, make sure that the element it is tied to appears in all these cases.

Secondly, this approach isolates each module in its own closure. Hopefully putting things behind the element check means everything always works, but if there are problems then errors in each closure won't break the rest of the page, so we only lose one module at a time.
In this case, pages that don't include this auto logout switch element will fail when we try and add properties and events to the element. This block will fail, and because it isn't isolated, it will stop executing this whole codeblock (which means potentially breaking all the modules on the page).

adamnfish: Actually, I thought I'd check (it has been a while) and I think this will be fine in practice - it'll be ok trying to do things to an empty element selection. I'll leave this comment as a description of why things are laid out the way they are though, you might still choose to put the setup behind a check that the `#auto_login_switch` element exists for consistency but up to you!

public/javascripts/janus.js
```diff
@@ -279,4 +279,30 @@ jQuery(function($){
...
+            localStorage.setItem(LOCAL_STORAGE_KEY__AUTO_LOGOUT, event.target.checked.toString());
+        };
+
+        $("a[href*='/console?permissionId=']").each(function(_, el){
+            el.onclick = (clickEvent) => {
+                if(autoLogoutSwitchElement.checked) {
+                    clickEvent.preventDefault();
+                    const targetHref = el.href;
+                    console.log("Silently attempting logout before navigating to", targetHref)
+                    fetch("https://signin.aws.amazon.com/logout", {
```
adamnfish, line 297: Should we put a timeout on this, so that we can be pretty sure the user's intent will be followed at some point?
I worry network issues or unusual browser settings (now or in the future) could leave Janus broken if they mean this call doesn't complete?

twrichards: great point, fixed in https://github.com/guardian/janus-app/pull/439/files/065b979f76e173b8a555342d726aa87e17e40910..c8a34438bb9739b61fbe69955b99443ce9fecf69

twrichards:
@adamnfish I've addressed all the comments you've raised - please could I get a further review?

adamnfish (approved):
Let's release this and see how we go.

One way to make the UX a bit tighter would be to give some feedback while the logout is in progress, and ideally disable the console link while the request is in flight. Not a huge deal, definitely something we can think about after we've proved this is useful!

Another potential issue that's more fundamental to the approach is that because this hacks the way the link works, it won't work if people do anything apart from click the link (e.g. open the page in a new tab or drag it to another window). I'm always nervous about providing functionality that is inconsistent like this, but we've chatted about this before and I understand that this implementation will be useful for you and others as is.

Thank you for contributing!

PR #444, opened 2024-06-05
Ignore failures in the test-reporter build step

adamnfish:
The test reporter step does not work for PRs raised from forks. We don't mind about this, it's a nice-to-have!

Let's ignore those failures.

This means contributions from outside DevX won't have their builds fail at the test-reporter step.

### DIFF
--- a/.github/workflows/trigger-private-janus-build.yml
+++ b/.github/workflows/trigger-private-janus-build.yml
@@ -34,6 +34,7 @@ jobs:
       - name: Test Report for Janus-App
         uses: dorny/test-reporter@31a54ee7ebcacc03a09ea97a7e5465a47b84aea5 # v1.9.1
         if: success() || failure() # run this step even if previous step failed
+        continue-on-error: true # this step fails when PRs are raised from forks, we don't mind!
         with:
           name: Janus-App Tests
           path: logs/test-reports/TEST-*.xml


.github/workflows/trigger-private-janus-build.yml
```diff
@@ -34,6 +34,7 @@ jobs:
       - name: Test Report for Janus-App
         uses: dorny/test-reporter@31a54ee7ebcacc03a09ea97a7e5465a47b84aea5 # v1.9.1
         if: success() || failure() # run this step even if previous step failed
```
AshCorr, line 36: Might it make sense to skip the report if its run on a fork?

```suggestion
        if: (success() || failure()) && ${{ !github.event.pull_request.head.repo.fork }} # run this step even if previous step failed
```

from https://til.cazzulino.com/devops-ci-cd/how-to-skip-steps-or-jobs-in-github-actions-for-prs-from-forks

AshCorr (approved):

.github/workflows/trigger-private-janus-build.yml
adamnfish: I think you're right this is better, because it is more explicit. This step is optional, but probably still better to know when it fails 👍

twrichards: note the $${{ }} are not needed here, and in fact mean it doesn't detect forks as desired, fixed in https://github.com/guardian/janus-app/pull/439/commits/a63dde8f88994472903a6ab27484bf1bd2d2fc95

PR #447, opened 2024-06-06
add an 'auto-logout toggle' (take 2)

twrichards:
This **reattempts #439**, which was reverted in #445 because it doesn't work in FireFox [and Chrome in a few months] owing to [stricter policy w.r.t. SameSite attribute of cookies](https://hacks.mozilla.org/2020/08/changes-to-samesite-cookie-behavior/) and AWS cookies need to be present when hitting logout. This time we achieve the same end result with [`window.open(   )`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) which is displayed for 0.75s before being closed and redirected (which should allow enough time for the logout to complete)...
janus_auto_logout_popup

### DIFF
--- a/public/javascripts/janus.js
+++ b/public/javascripts/janus.js
@@ -279,4 +279,33 @@ jQuery(function($){
         }
     });
 
+    // auto-logout (preference persisted with local storage)
+    $("#auto_logout_switch").each(function(_, autoLogoutSwitchElement){
+        const LOCAL_STORAGE_KEY__AUTO_LOGOUT = "autoLogout"
+        autoLogoutSwitchElement.checked = localStorage.getItem(LOCAL_STORAGE_KEY__AUTO_LOGOUT) === "true";
+        autoLogoutSwitchElement.onchange = (event) => {
+            autoLogoutSwitchElement.checked = event.target.checked;
+            localStorage.setItem(LOCAL_STORAGE_KEY__AUTO_LOGOUT, event.target.checked.toString());
+        };
+
+        $("a[href*='/console?permissionId=']").each(function(_, el){
+            el.onclick = (clickEvent) => {
+                if(autoLogoutSwitchElement.checked) {
+                    clickEvent.preventDefault();
+                    const targetHref = el.href;
+                    console.log("Attempting logout before navigating to", targetHref)
+                    const popup = window.open(
+                      "https://signin.aws.amazon.com/logout",
+                      "AWSLogoutPopup",
+                      "width=500, height=500, top=100, left=100, popup=true"
+                    );
+                    setTimeout(() => {
+                        popup.close();
+                        location.href = targetHref;
+                    }, 750);
+                }
+            }
+        });
+    });
+
 });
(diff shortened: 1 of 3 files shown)


public/javascripts/janus.js
```diff
@@ -279,4 +279,33 @@ jQuery(function($){
...
+                    console.log("Attempting logout before navigating to", targetHref)
+                    const popup = window.open(
+                      "https://signin.aws.amazon.com/logout",
+                      "AWSLogoutPopup",
+                      "width=500, height=500, top=100, left=100, popup=true"
+                    );
+                    setTimeout(() => {
+                        popup.close();
+                        location.href = targetHref;
+                    }, 750);
```
adamnfish, line 305: If we can't detect a successful logout request (which as we've discussed, really ought not to be possible) then there's an inherent trade-off between inconvenience and reliability. 750ms is aaaages so it should be fine, but we know it isn't always going to be. We're delaying the page by enough to be annoying, without it being enough to be reliable.

This is on top of the other issues with the approach. For example, I usually open links in a new tab and this feature won't work in that case. We're building something that doesn't really fit with how web UIs work, which means it is sure to have surprising behaviour from time to time. This is why I went down a quick rabbit hole on this feature a few years back, and we ended up [deciding to wait and see if AWS offer support](https://github.com/guardian/janus/issues/868#issuecomment-313160855) / people needed it enough.

But this PR suggests people do want to the feature! It's always a delight to get a contribution from someone that wants to improve the tool for themselves and their colleagues. 750ms is way less than doing it manually, I can see how this will be useful to people. so if this is a feature people want we should go ahead with it.

A word of caution though, thesee UI papercuts all add up. UIs are the main way people interact with our tools and unreliable UIs add to the sense that our tooling is old, poorly maintained, or no longer fit for purpose. We should aspire to build UIs that are reliable. They should be accessble, responsive (to interaction and browser size), elegantly handle edge cases, give good error feedback, and make impossible states impossible. By far the easiest way to achieve this is to build standards-compliant UIs that lean into the browser platform - not reinventing browser feautres or (in this case) trying to work around them.

jamesgorrie: I really like this the outline of what is in tension here!

I do agree with the sentiment of being able to achieve super-slick UIs, especially ones we use continuously through the day. Wasting time on substandard UIs ultimately means I use them less or it just makes my day feel like mud.

But - the current situation for some engineers is muddier than mud[^1] in regards to the current logout workflow - and this alleviates some of the faff. Could this be v0.1 of the feature, get feedback on it e.g. How often it fails, do people find it useful, do people even use it etc etc. If people like it - it could indicate needing to iterate on it. It might even prompt someone who then gets frustrating with this trade-off to try find a better way? If it's not really used, or deemed not easy enough to use - it might not worth the overhead and clutter - and we can remove it.

If we do choose to try it - I also wonder if we could be a little more optimistic and go for < 500ms? 750ms is aaaaages, or at least a noticeable amount. The failure in that instance then s you get the old logout flow 🤷.


[^1]: I'll drop the analogy here

PR #448, opened 2024-06-07
add an 'auto-logout toggle' (take 3 - proper way)

twrichards:
Looking to achieve the same as #439 and #447 were trying to do, but achieving it server-side, if toggle is active (persisted via cookie now, so server can see the preference on the request) [for AWS Console buttons] the server redirects to `https://us-east-1.signin.aws.amazon.com/oauth?Action=logout` with a `redirect_uri` query param containing the federated login link Janus would normally redirect to directly (see https://serverfault.com/a/1097528).

image

**✅ TESTED the logout with `redirect_uri` approach manually** in the browser console... 
```javascript
location.href=`https://us-east-1.signin.aws.amazon.com/oauth?Action=logout&redirect_uri=${encodeURIComponent("https://us-east-1.signin.aws.amazon.com/federation?Action=login&SigninToken=REDACTED&Issuer=https%3A%2F%2Fjanus.gutools.co.uk&Destination=https%3A%2F%2Fconsole.aws.amazon.com%2F")}`
```
... obviously with `REDACTED` being a real token (which I got by observing/copying from the `federation` link via the Network tab of browser developer tools)

### DIFF
--- a/app/controllers/Janus.scala
+++ b/app/controllers/Janus.scala
@@ -13,6 +13,8 @@ import org.joda.time.{DateTime, DateTimeZone, Duration}
 import play.api.{Configuration, Logging}
 import play.api.mvc._
 
+import java.net.URLEncoder
+
 class Janus(
     janusData: JanusData,
     controllerComponents: ControllerComponents,
@@ -98,7 +100,20 @@ class Janus(
       )
       loginUrl = Federation.loginUrl(credentials, host, stsClient)
     } yield {
-      SeeOther(loginUrl)
+      val redirectUrl = request.cookies.get("autoLogout") match {
+        case Some(cookie) if cookie.value == "true" =>
+          // NOTE us-east-1 is required in these URLs, as per https://serverfault.com/questions/985255/1097528#comment1469112_1097528
+          s"https://us-east-1.signin.aws.amazon.com/oauth?Action=logout&redirect_uri=${URLEncoder.encode(
+              loginUrl.replace(
+                "https://signin.aws.amazon.com",
+                "https://us-east-1.signin.aws.amazon.com"
+              ),
+              "UTF-8"
+            )}"
+        case _ =>
+          loginUrl
+      }
+      SeeOther(redirectUrl)
         .withHeaders(CACHE_CONTROL -> "no-cache")
     }) getOrElse {
       logger.warn(
--- a/public/javascripts/janus.js
+++ b/public/javascripts/janus.js
@@ -279,4 +279,16 @@ jQuery(function($){
         }
     });
 
+    // auto-logout (preference persisted via cookie, so server-side can see it when redirecting to federation endpoint)
+    $("#auto_logout_switch").each(function(_, autoLogoutSwitchElement){
+        const COOKIE__AUTO_LOGOUT = "autoLogout"
+        autoLogoutSwitchElement.checked =
+          !!decodeURIComponent(document.cookie)
+          .split(";")
+          .find(_ => _.trim().startsWith(`${COOKIE__AUTO_LOGOUT}=true`));
+        autoLogoutSwitchElement.onchange = (event) => {
+            document.cookie = `${COOKIE__AUTO_LOGOUT}=${event.target.checked}; expires=Fri, 31 Dec 9999 23:59:59 GMT; path=/`
+        };
+    });
+
 });
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -13,12 +13,15 @@
 
         <div class="row">
             <div class="logout-button">
-            <a href="https://signin.aws.amazon.com/oauth?Action=logout"
-               target="_blank"
-               class="waves-effect waves-light btn">
-                <i class="material-icons">exit_to_app</i>
-                logout
-            </a>
+                <span class="switch" title="Automatically logout before entering a new account's console">
+                    <label><span>Auto-logout</span><input id="auto_logout_switch" type="checkbox"><span class="lever"></span></label>
+                </span>
+                <a href="https://signin.aws.amazon.com/oauth?Action=logout"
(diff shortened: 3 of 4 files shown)


app/controllers/Janus.scala
```diff
@@ -98,7 +100,20 @@ class Janus(
       )
       loginUrl = Federation.loginUrl(credentials, host, stsClient)
     } yield {
-      SeeOther(loginUrl)
+      val redirectUrl = request.cookies.get("autoLogout") match {
```
adamnfish, line 103: I'd love to see this logic be tested, which would also give us the opportunity to enforce some of the things you've discovered (liek the us-east-1 constrain). Could this logic move to the `logic` package, perhaps we can create a new object in there if there isn't a good existing candidate already. With this as a function we can write some tests that takes the "state" (currently from the cookie but see below) and confirm that we have returned the correct URL. If we do that, let's focus on the properties we're checking with a helpfully named test for each? I'll be happy to look at this with you if you'd like?

Additionally, I wonder whether this logic could be part of the work `Federation.loginUrl` achieves. That function returns the login URL, so it might make sense for it to take the "logout state" and return the correct URL based on that? This function performs side-effects (Amazon provide the signed console link from the credentials) so we'd still want this logic to be available elsewhere and tested, it's just a question of where it gets called. I think having the `loginUrl` on [line 101](https://github.com/guardian/janus-app/pull/448/files#diff-7ebc3def08c39769f2b0e152b9891084e6be013ffc4de7d8ea164fa571b845fbR101) be the answer from this controller's point of view would be a little clearer, but let me know what you think!

public/javascripts/janus.js
```diff
@@ -279,4 +279,16 @@ jQuery(function($){
...
 
+    // auto-logout (preference persisted via cookie, so server-side can see it when redirecting to federation endpoint)
+    $("#auto_logout_switch").each(function(_, autoLogoutSwitchElement){
+        const COOKIE__AUTO_LOGOUT = "autoLogout"
+        autoLogoutSwitchElement.checked =
+          !!decodeURIComponent(document.cookie)
+          .split(";")
+          .find(_ => _.trim().startsWith(`${COOKIE__AUTO_LOGOUT}=true`));
+        autoLogoutSwitchElement.onchange = (event) => {
+            document.cookie = `${COOKIE__AUTO_LOGOUT}=${event.target.checked}; expires=Fri, 31 Dec 9999 23:59:59 GMT; path=/`
```
adamnfish, line 290: Would it be simpler to add a parameter to the `consoleLogin` action, rather than have us synchronise the reading and writing of this cookie? We could let the UI choose the most effective way to collect and persist the user's preference, and have that be communicated explicitly when it talks to the server?

In this case, that might mean sticking to localStorage for the preference and including a querystring parameter or optional path parameter to request a "logout-first" console URL?

Side note: The "favourites" feature does use a cookie so there is some prioir art here! However, this is because the favourites feature needs to change the way the homepage renders on the server. The favourites feature is inherently a piece of server-state, so the choice there is infra-side state via a database or to have that state be client-persisted-server-available via a cookie.

adamnfish (commented):
I think this is a great approach. Really nice catch that this becomes possible if we stick to using us-east-1, that's such a weird detail that unlocks doing this in a very sensible way! 🤪 

I've added a few comments about the code, happy to chat them over with you! The only other thing is that it'd be gerat to double check that we're good to use the redirect_uri parameter for this kind of secret information. Generally we see the world moving towards using Headers instead of QS params for this kind of sensitive information. I think this is the established way to do console login type things with AWS (the `getSigninToken` stuff works a similar way already) so I expect it's fine, but let's chat to AWS to make sure before merging this. I'll want to hear that AWS themselves are treating this parameter the same way they'd treat API keys in general.

twrichards:
> it'd be gerat to double check that we're good to use the redirect_uri parameter for this kind of secret information. Generally we see the world moving towards using Headers instead of QS params for this kind of sensitive information. I think this is the established way to do console login type things with AWS (the `getSigninToken` stuff works a similar way already) so I expect it's fine, but let's chat to AWS to make sure before merging this. I'll want to hear that AWS themselves are treating this parameter the same way they'd treat API keys in general.

Yeah I agree, although there's no guarantee AWS won't change their internals and start logging it where they shouldn't (if perhaps a team were interested in where people were redirecting too, I could picture it coming out on a dashboard somewhere)... according [to the docs they support POST](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_enable-console-custom-url.html#STSConsoleLink_manual_step5) 
image
... and by this I presume they mean you can send the params in the body (`application/x-www-form-urlencoded`), can't picture a way that this helps us though.

twrichards:
Commenting to remove stale label, since apparently I can't the normal way. @adamnfish I don't recall seeing any reply on support case. I'm not working for 6 weeks, I know you said you might add some commits on this branch/PR?

public/javascripts/janus.js
adamnfish: In the replacement PR I've decided not to take the approach of switching away from cookies. The key reason is that I think we should respect this setting for existing URLs (e.g. those that are bookmarked). This is particularly obvious with the "console url" feature. This feature as-written would have hard-coded the logout decision in the provided URL, which may not be what users expect.

app/controllers/Janus.scala
adamnfish: In the PR that replaces this one, I've addressed this point by pulling out the redirect logic into a tested function. This also gives us the opportunity to add the auto-logout functionality to both the "console URL" and "console redirect" features without needing to duplicate the logic.

adamnfish:
I've duplicated this PR in the janus-app repository, because I don't have access to this fork to easily make additions.

https://github.com/guardian/janus-app/pull/456

PR #459, opened 2024-08-23
Fix the DB for local development

adamnfish:

Fixes the database when running the application locally.

## More detail

The helper library being used for DB access is not compatible with recent versions of DynamoDB loal. DynamoDB now verifies that access credentials have the right format, even for local development where they are not actually used. The library was using empty strings for the access key and secret, which did not pass validation.

This PR adds a utility for creating a valid DynamoDB client for local use, which is consumed from the server itself (in DEV mode) and the database integration tests.

## Deprecation suppression

Unfortunately, the code that sets up the local DB client also uses a deprecated method, so we suppress this deprecation warning for now. Fixing this properly will mean moving off AWScala, a job that we already know about.

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -2,7 +2,6 @@ import aws.Clients
 import awscala.dynamodbv2.DynamoDB
 import com.amazonaws.regions.Regions
 import com.gu.googleauth.AuthAction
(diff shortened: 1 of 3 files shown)


tjsilver (approved):
👍

PR #479, opened 2024-11-20
document AwsAccount case class

NovemberTang:

When [trying to add a new account to janus](https://github.com/guardian/janus/pull/4447), I clicked through to the definition of AwsAccount to work out what the parameters were. Looking at the definition of this case class from there, it wasn't immediately obvious what they were supposed to mean/represent, so I've added a bit of documentation to clarify this for future users.

I'm not sure why it's called `authConfigKey`, but if there's no specific reason, we could maybe eliminate the second comment by renaming that param to `profileName`? Interested to hear people's thoughts.

It's now more immediately clear what the parameters represent.

### DIFF
--- a/configTools/src/main/scala/com/gu/janus/model/models.scala
+++ b/configTools/src/main/scala/com/gu/janus/model/models.scala
@@ -42,7 +42,11 @@ object SupportACL {
 }
 
 case class AwsAccount(
+    /* The name of the account as you would like it to appear on Janus */
     name: String,
+    /** The name of the AWS profile to pass in via CLI. Should be relatively
+      * short, with no spaces
+      */
     authConfigKey: String
 )
 case class AwsAccountAccess(


configTools/src/main/scala/com/gu/janus/model/models.scala
```diff
@@ -42,7 +42,11 @@ object SupportACL {
 }
 
 case class AwsAccount(
+    /* The name of the account as you would like it to appear on Janus */
     name: String,
+    /** The name of the AWS profile to pass in via CLI. Should be relatively
```
adamnfish, line 47: As mentioned in the PR comment, this string is actually used for three separate purposes. I think it'd be nice if this comment conveyed that information so people are (hopefully) less confused in the future?

Do you feel like the comment on the PR gives you enough context to understand this case class field, and write a comment that conveys that understanding? If not then let me know, and we can look together!

I'd also consider refactoring this into something clearer at some point, but we'll be spending a bit of time on Janus soon so that's probably a good time for us to iron out some of these kinds of kinks.

adamnfish (commented):
> I'm not sure why it's called authConfigKey

It's called the auth config key because it's the key for looking up the auth in the config. More specifically, which role to assume to get the access Janus needs to bestow access to the chosen AWS account.

I think this would be fine, but unfortunately, this string is also used for two other purposes and both of those are much more visible to users:

1. the (default) profile name for created janus CLI sessions in that account (assuming you use the standard mechanism of populating an AWS profile)
2. part of the permission key used to uniquely identify a permission request (you'll see this in the URL of Janus console / credentials requests)

It's a classic case of an API that evolved while the app was created, and then never had a second pass to tidy things up.

(p.s. the failed build is a lint error, running scalafmt on the code should fix that, let me know if you'd like to look with me?)

adamnfish:
I've recreated this PR with my suggested changes on a separate branch so that we can get this released and merged. Thank you for the feedback, and if you get a moment to reconsider this change please let us know if the new wording is clear enough!

https://github.com/guardian/janus-app/pull/487

PR #487, opened 2024-12-23
Document the AWS Account model fields

adamnfish:
When people set up a new AWS account in Janus it helps if people clearly understand what these fields are for. Because the `authConfigKey` field has multiple purposes, it helps to make this clear to the reader.

It should be easier for people to onboard new AWS accounts in the future.

## Any additional notes?

Replaces https://github.com/guardian/janus-app/pull/479 in a fork that the current DevX/Security team members control.


tjsilver (approved):
Very helpful 👍

PR #491, opened 2025-01-09
Switch to guardian/setup-scala in GHA workflows

adamnfish:
This workflow installs Java and sbt, and is managed by us. Also updates the .tool-versions file, which was pointing at an out-of-date version of Java. We now only specify the major Java version, which is supported locally by `mise` (but not `asdf`).

### DIFF
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -11,15 +11,8 @@ jobs:
       - name: Checkout branch
         id: checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-      - name: Install Java
-        id: java
-        uses: actions/setup-java@7a6d8a8234af8eb26422e24e3006232cccaa061b # v4.6.0
-        with: 
-          distribution: 'corretto'
-          java-version: '17'
-      - name: Install SBT
-        id: install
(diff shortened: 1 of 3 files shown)


kelvin-chappell:
Should this also be updated for consistency? [-release:11](https://github.com/guardian/janus-app/blob/main/build.sbt#L32)  
Not strictly necessary.

The release workflow also uses the tool-versions file.  Not sure whether that assumes an asdf format.

adamnfish:
> Should this also be updated for consistency? [-release:11](https://github.com/guardian/janus-app/blob/main/build.sbt#L32) Not strictly necessary.

I don't think so! Because we're publishing a library, we should make sure it is as widely consumable as possible, which means setting a lower JDK version as the target. We generally want to use the highest sensible version ourselves (I think that's 21 right now for Scala), but target the lowest sensible version when publishing (I think that's currently 11).

> The release workflow also uses the tool-versions file. It says that [it needs an asdf format](https://github.com/guardian/gha-scala-library-release-workflow/blob/1cb253ef292b3c1d975a8f848d1464b290792d9b/.github/workflows/reusable-release.yml#L116). But actually the [logic to find major version](https://github.com/guardian/gha-scala-library-release-workflow/blob/1cb253ef292b3c1d975a8f848d1464b290792d9b/.github/workflows/reusable-release.yml#L119) will work with either format.

Great shout, and great to check, thank you!

kelvin-chappell (approved):
👍

PR #495, opened 2025-01-30
Add public information about local development on audit endpoints

kelvin-chappell:
This fills a small gap in the documentation on local development.


local-dev/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+To make this table available for testing auditing functionality:
+
+1. Start Docker
+2. Start up a local Dynamo DB container:
+```shell
+docker-compose up -d
+```
+3. Create the audit table schema by running [Creation test](test/aws/AuditTrailDBTest.scala#L46).
+4. Populate the table with some test data by running [Insertion test](test/aws/AuditTrailDBTest.scala#L52).
```
adamnfish, line 15: I think the link we'd want to give people is the following:
https://github.com/guardian/janus-app/blob/main/test/aws/AuditTrailDBTest.scala#L46-L48

This is the test that sets up the database collections, a requirement for running it locally.

local-dev/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      - dynamodata:/home/dynamodblocal
+    working_dir: /home/dynamodblocal
+    command: "-jar DynamoDBLocal.jar -sharedDb -dbPath ."
+
+networks:
+  local-dynamodb:
+    name: local-dynamodb
+
+volumes:
+  dynamodata: {}
```
adamnfish, line 19: Is it worth mounting the shared DB under `~/.gu/janus-app` on the host? I don't feel strongly, but would mean that you don't have to recreate the db tables any time you update the container.

adamnfish (commented):
This used to be documented using the old "download the jar and run it" approach, great to add the proper modern (docker) way, thank you!

Can we say a little more about having to create the database collections as a one-shot the first time you do this? We could automate this in the docker-compose file at some point, but for now I like the approach to just document it.

Is it worth adding a note (/link) to the "running locally" docs in our janus configuration repo as well?

local-dev/docker-compose.yml
kelvin-chappell: Yes that's a good idea

adamnfish (approved):
I didn't mean for the local mount / persistence to be a blocking change, sorry! Let's get this merged and we can follow up with a tweak.

> Can we say a little more about having to create the database collections as a one-shot the first time you do this? We could automate this in the docker-compose file at some point, but for now I like the approach to just document it.

For this as well, I think it'd be better to move this to a `main` method somewhere, rather than having to fiddle with tests. That's an easy health tweak at some and these docs are great as-is, thank you!

kelvin-chappell:
> I didn't mean for the local mount / persistence to be a blocking change, sorry! Let's get this merged and we can follow up with a tweak.

No problem - I just didn't prioritise it.  I'll make the two fixes now and then get this out.

local-dev/README.md
kelvin-chappell: Fixed the line refs in 5ede52e8e910b401e68a52e640c83e30bc75f10b

local-dev/docker-compose.yml
kelvin-chappell: Done in 5ed9d687a0c1feef1877e22188b4e7db6d1b7585

PR #507, opened 2025-02-11
Remove logout to support aws multiple console sessions

adamnfish:

Makes Janus compatible with the new multi-login support in the AWS console. The existing logout (and auto-logout) functionality exists to support users' ability to easily switch between AWS accounts with Janus, but the ability to crteate multiple concurrent AWS console logins means this is no longer required.

We now encourage users to turn on AWS' multi-session support, and remove the now redundant (and broken) login functionality from Janus. Users obtain AWS access from Janus as before, and can now manage their sessions directly from the AWS console via the drop-down at the top right.

Our colleagues will be able to use Janus to seamlessly log into multiple AWS accounts.

## Images

This shows how to enable multi-session support in the AWS console, using the drop-down in the top right of the AWS Console. Users need to individually enable this for themselves:

enable-aws-multi-session

and here it is in action:

aws-multi-accounts

Users can now switch between sessions, and log out of all/specific sessions from the AWS console.

## Any additional notes?

I've already tested that the multi-session functionality works without auto-logout enabled, this change tjust removes that feature. We'll check this carefully post-deploy, and roll back if required.

### DIFF
--- a/app/aws/Federation.scala
+++ b/app/aws/Federation.scala
@@ -93,42 +93,13 @@ object Federation {
   def generateLoginUrl(
       temporaryCredentials: TemporaryCredentials,
       host: String,
-      autoLogout: Boolean,
       sts: STS
   ): String = {
-    val url = sts.loginUrl(
+    sts.loginUrl(
       credentials = temporaryCredentials,
       consoleUrl = "https://console.aws.amazon.com/",
       issuerUrl = host
     )
-    autoLogoutUrl(url, autoLogout)
-  }
-
-  /** Janus supports logging users out before redirecting them to the Console.
-    *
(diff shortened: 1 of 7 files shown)


tjsilver (approved):
👍 Tested locally and observed the button and toggle are absent.

adamnfish:
Thank you! Before we merge this I am going to:

- check that the behaviour is sensible when a Janus session expires
- draft some comms for the department explaining about this new feature and the deprecation of the logout button

adamnfish:
> check that the behaviour is sensible when a Janus session expires

Works great!

I've drafted comms and shared them separately, so we can roll this out when we've some time!

PR #509, opened 2025-02-13
Upgrade AWS IAM v1 client to v2

kelvin-chappell:
This change removes all dependence on [AWS v1 SDKs](https://github.com/aws/aws-sdk-java) and on the [archived awscala library](https://github.com/seratch/AWScala).
All interaction with AWS is now done through [v2 SDKs](https://github.com/aws/aws-sdk-java-v2/).

This will make the code more secure and easier to maintain.

## Any additional notes?
AWS provides a [library for modelling IAM policies](https://mvnrepository.com/artifact/software.amazon.awssdk/iam-policy-builder) but to make use of this library would involve changing a lot of code in this repo and in the call sites of the [configTools library](https://github.com/guardian/janus-app/tree/fd3f01d5db3f2d407b168438da15d381988fa807/configTools/src), making the change history difficult to navigate.
Instead we have created case classes to model policies that duplicate the pre-existing awscala case classes.
This has the benefit of only needing to change imports in most cases so preserving the change history.  But it comes at the cost of having to generate a json representation of policies, which is more complicated than just auto-encoding the case class structure.

## To test
Verify in production that:

- [ ] Can still sign into AWS console from Janus
- [ ] Can sign out from console and back in again from Janus
- [ ] Can retrieve credentials and use them in command line locally
- [ ] Can revoke all credentials for an account and see policy condition has correct threshold time and console and local credentials unusable
- [ ] Logs have no new errors

## See also
* #500
* #504

### DIFF
--- a/app/aws/Federation.scala
+++ b/app/aws/Federation.scala
@@ -1,18 +1,15 @@
 package aws
 
-import awscala.iam._
-import com.amazonaws.auth.AWSStaticCredentialsProvider
-import com.amazonaws.services.identitymanagement.model.GetRoleRequest
 import com.gu.janus.model.{AwsAccount, Permission}
 import data.Policies
 import logic.Date
 import org.joda.time.{DateTime, DateTimeZone, Duration, Period}
 import play.api.libs.json.Json
+import software.amazon.awssdk.auth.credentials._
+import software.amazon.awssdk.services.iam.IamClient
+import software.amazon.awssdk.services.iam.model.PutRolePolicyRequest
 import software.amazon.awssdk.services.sts.StsClient
-import software.amazon.awssdk.services.sts.model.{
-  AssumeRoleRequest,
-  Credentials
-}
+import software.amazon.awssdk.services.sts.model._
 
 import java.net.{URI, URLEncoder}
 import java.nio.charset.StandardCharsets.UTF_8
@@ -141,24 +138,24 @@ object Federation {
       stsClient,
       Federation.awsMinimumSessionLength
     )
-    val sessionCredentials = awscala.Credentials(
+    val sessionCredentials = AwsSessionCredentials.create(
       creds.accessKeyId,
       creds.secretAccessKey,
       creds.sessionToken
     )
-    val provider = new AWSStaticCredentialsProvider(sessionCredentials)
-    val iamClient = IAM(provider)
+    val provider = StaticCredentialsProvider.create(sessionCredentials)
+    val iamClient = IamClient.builder().credentialsProvider(provider).build()
 
     // remove access from assumed role
     val roleName = getRoleName(roleArn)
-    val getRoleRequest = new GetRoleRequest().withRoleName(roleName)
-    val role = Role(iamClient.getRole(getRoleRequest).getRole)
-    val roleRevocationPolicy =
-      RolePolicy(role, "janus-role-revocation-policy", revocationPolicyDocument)
-    // ^
-    // this name should match policy in cloudformation/federation.template.yaml
+    val roleRevocationPolicy = PutRolePolicyRequest
+      .builder()
+      .roleName(roleName)
+      .policyName("janus-role-revocation-policy")
+      .policyDocument(revocationPolicyDocument)
+      .build()
     iamClient.putRolePolicy(roleRevocationPolicy)
-    iamClient.shutdown()
+    iamClient.close()
   }
 
   private[aws] def getRoleName(roleArn: String): String = {
--- a/app/aws/Federation.scala
+++ b/app/aws/Federation.scala
@@ -152,6 +152,8 @@ object Federation {
       .builder()
       .roleName(roleName)
       .policyName("janus-role-revocation-policy")
+      //           ^
+      // this name should match policy in cloudformation/federation.template.yaml
       .policyDocument(revocationPolicyDocument)
       .build()
     iamClient.putRolePolicy(roleRevocationPolicy)
--- a/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
+++ b/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
@@ -73,12 +73,12 @@ object Iam {
         .groupBy(_.typeName)
         .view
         .mapValues(conditions =>
-          Json.obj(
+          Json.fromFields(
             conditions.map(condition =>
               condition.key -> Json.fromValues(
                 condition.conditionValues.map(Json.fromString)
               )
-            ): _*
+            )
           )
         )
         .toMap
--- a/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
+++ b/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
@@ -40,13 +40,6 @@ object Iam {
   )
 
   case class Principal(id: String, provider: String)
-  object Principal {
-    implicit val encoder: Encoder[Principal] = Encoder.instance { principal =>
-      Json.obj(
-        principal.provider -> Json.fromString(principal.id)
-      )
-    }
-  }
 
   case class Statement(
       effect: Effect,
--- a/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
+++ b/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
@@ -38,17 +38,6 @@ object Iam {
       typeName: String,
       conditionValues: Seq[String]
   )
-  object Condition {
-    implicit val encoder: Encoder[Condition] = Encoder.instance { condition =>
-      Json.obj(
-        condition.typeName -> Json.obj(
-          condition.key -> Json.fromValues(
-            condition.conditionValues.map(Json.fromString)
-          )
-        )
-      )
-    }
-  }
 
   case class Principal(id: String, provider: String)
   object Principal {
--- a/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
+++ b/configTools/src/main/scala/com/gu/janus/policy/Iam.scala
@@ -1,0 +1,148 @@
+package com.gu.janus.policy
+
+import io.circe.syntax._
+import io.circe.{Encoder, Json, JsonObject}
+
+import scala.collection.immutable.ListMap
+import scala.collection.mutable
+
+/** This models the AWS IAM policy types and subtypes.
+  *
+  * See https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies.html
+  * for the logic of the Json encoding.
+  */
+object Iam {
+
+  sealed trait Effect
+  object Effect {
+    case object Allow extends Effect
+    case object Deny extends Effect
+
+    implicit val encoder: Encoder[Effect] = Encoder.instance {
+      case Allow => Json.fromString("Allow")
+      case Deny  => Json.fromString("Deny")
+    }
+  }
+
+  case class Action(name: String)
+  object Action {
+    implicit val encoder: Encoder[Action] = Encoder[String].contramap(_.name)
+  }
+
+  case class Resource(id: String)
+  object Resource {
+    implicit val encoder: Encoder[Resource] = Encoder[String].contramap(_.id)
+  }
+
+  case class Condition(
+      key: String,
+      typeName: String,
+      conditionValues: Seq[String]
+  )
+  object Condition {
+    implicit val encoder: Encoder[Condition] = Encoder.instance { condition =>
+      Json.obj(
+        condition.typeName -> Json.obj(
+          condition.key -> Json.fromValues(
+            condition.conditionValues.map(Json.fromString)
+          )
+        )
+      )
+    }
+  }
+
+  case class Principal(id: String, provider: String)
+  object Principal {
+    implicit val encoder: Encoder[Principal] = Encoder.instance { principal =>
+      Json.obj(
+        principal.provider -> Json.fromString(principal.id)
+      )
+    }
+  }
+
+  case class Statement(
+      effect: Effect,
+      actions: Seq[Action],
+      resources: Seq[Resource],
+      id: Option[String] = None,
+      conditions: Seq[Condition] = Nil,
+      principals: Seq[Principal] = Nil
+  )
+  object Statement {
+    private def conditionsAsJson(conditions: Seq[Condition]): Json = {
+      val mergedConditions = conditions
+        .groupBy(_.typeName)
+        .view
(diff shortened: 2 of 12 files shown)


test/logic/UserAccessTest.scala
```diff
@@ -1,10 +1,9 @@
 package logic
 
-import awscala.DateTime
 import com.gu.googleauth.UserIdentity
 import fixtures.Fixtures._
 import com.gu.janus.model.{ACL, SupportACL}
-import org.joda.time.{DateTimeZone, Period}
+import org.joda.time.{DateTime, DateTimeZone, Period}
```
tjsilver, line 6: Should we use Java time now instead?

tjsilver (approved):
Have tested locally and all the expectations listed were met.

test/logic/UserAccessTest.scala
kelvin-chappell: Yes we should!  I'm working on that on another branch

adamnfish (commented):
Looks good from a reading the code point of view! I haven't yet actually exercised these bits locally.

As well as testing that the Janus webapp still works, it'd be great to know for sure we aren't affecting the Janus Data config format that will be produced by consumers of the config-tools library. One way to do that would be to take an example configuration generated both before and after this change. We can (privately) use our real Guardian config of course, but another candidate is the "example" project in this repository. This sub-project doesn't get built in CI/CD and I think should probably go away completely, but while it exists maybe that can serve as a useful sample config so we can check in a reference file during this transition period.

I think the config file serialisation behaviour is as key a guarantee for us to preserve as Janus' actual end-user behaviour is. We'd like the existing and new Janus Data files to be readable by the existing and new versions of the code, so that the migration (and reversion) path is as smooth as possible. Whenever we have an important property of the code like this, ideally we have that captured in a test!

Note: heads-up that it looks like comparing the current and previous Janus Data files will result in statement ID differences, as I mention below.

configTools/src/main/scala/com/gu/janus/policy/Iam.scala
```diff
@@ -0,0 +1,147 @@
...
+import io.circe.{Encoder, Json, JsonObject}
+
+import scala.collection.immutable.ListMap
+
+/** This models the AWS IAM policy types and subtypes.
+  *
+  * See https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies.html
+  * for the logic of the Json encoding.
+  */
+object Iam {
```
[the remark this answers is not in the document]
kelvin-chappell: It seemed like a convenient reference to hold everything together.  But can remove if you prefer - not strictly necessary.

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
```diff
@@ -0,0 +1,209 @@
...
+          )
+        ),
+        id = Some("ComplexPolicy")
+      )
+
+      val json = complexPolicy.asJson.noSpaces
+      // Verify that the generated JSON is valid
+      decode[JsonObject](json).isRight shouldBe true
+
+      // Verify specific elements
```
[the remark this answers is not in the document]
kelvin-chappell: Good point.  I'll break down the tests and add a full json test.

configTools/src/main/scala/com/gu/janus/policy/Iam.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    implicit val encoder: Encoder[Statement] = Encoder.instance { statement =>
+      // Using ListMap to preserve order of fields, which is easier to debug
+      val baseFields = ListMap(
+        "Effect" -> statement.effect.asJson,
+        "Action" -> statement.actions.asJson,
+        "Resource" -> statement.resources.asJson
+      )
+
+      val withSid = statement.id.fold(baseFields)(id =>
```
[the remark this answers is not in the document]
kelvin-chappell: Now statement IDs will be included in the output json representation if they're explicitly added to a [Statement case class](https://github.com/guardian/janus-app/blob/24dfd12643f2c8e877a4d285ef7c0a72234b597b/configTools/src/main/scala/com/gu/janus/policy/Iam.scala#L66), otherwise they'll be omitted.

kelvin-chappell:
> > As well as testing that the Janus webapp still works, it'd be great to know for sure we aren't affecting the Janus Data config format that will be produced by consumers of the config-tools library. One way to do that would be to take an example configuration generated both before and after this change. We can (privately) use our real Guardian config of course, but another candidate is the "example" project in this repository. This sub-project doesn't get built in CI/CD and I think should probably go away completely, but while it exists maybe that can serve as a useful sample config so we can check in a reference file during this transition period.

> > I think the config file serialisation behaviour is as key a guarantee for us to preserve as Janus' actual end-user behaviour is. We'd like the existing and new Janus Data files to be readable by the existing and new versions of the code, so that the migration (and reversion) path is as smooth as possible. Whenever we have an important property of the code like this, ideally we have that captured in a test!

> > Note: heads-up that it looks like comparing the current and previous Janus Data files will result in statement ID differences, as I mention below.
> 
> I'll set this up and probably remove the statement IDs if they're redundant.

A comparison of the data file output from the example project using v1.0.0 of configTools against the latest snapshot version of configTools showed that the output was identical. (Once I had reformatted, removed statement IDs and manually reordered object attributes). ie. all keys are the same and json without statement IDs is identical.

configTools/src/main/scala/com/gu/janus/policy/Iam.scala
```diff
@@ -0,0 +1,147 @@
...
+      conditions: Seq[Condition] = Nil,
+      principals: Seq[Principal] = Nil
+  )
+  object Statement {
+    private def conditionsAsJson(conditions: Seq[Condition]): Json = {
+      val mergedConditions = conditions
+        .groupBy(_.typeName)
+        .view
+        .mapValues(conditions =>
+          Json.obj(
```
adamnfish, line 76: Would it be clearer to use `Json.fromFields` rather than passing the collection as varargs (`: _*`) to the `Json.obj` constructor?

configTools/src/main/scala/com/gu/janus/policy/Iam.scala
```diff
@@ -0,0 +1,147 @@
...
+    implicit val encoder: Encoder[Resource] = Encoder[String].contramap(_.id)
+  }
+
+  case class Condition(
+      key: String,
+      typeName: String,
+      conditionValues: Seq[String]
+  )
+  object Condition {
+    implicit val encoder: Encoder[Condition] = Encoder.instance { condition =>
```
adamnfish, line 42: Is this actually being used? (The same applies to the Principal below). I ask because [in the Statement serialisation](https://github.com/guardian/janus-app/pull/509/files#diff-70fa0cae455542b80f233a55798d6f87af322305b5deb9b4786cd84ade56db7eR93-R95) we duplicate this implementation rather than deferring to this serialisation.

I think either way round is fine (inlining it all into the statement serialisation, or separating the serialisation of each type into its own encoder), but I don't think we should leave unused encoders lying around so that people don't suffer the confusion of editing them without effect in the future?

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
```diff
@@ -0,0 +1,269 @@
+package com.gu.janus.policy
+
+import com.gu.janus.policy.Iam._
+import io.circe.JsonObject
+import io.circe.parser.decode
+import io.circe.syntax._
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.should.Matchers
+
+class IamSpec extends AnyFreeSpec with Matchers {
```
adamnfish, line 10: All the other tests use the `XXTest` naming convention (rather than `XXSpec`). Minor point, but it might make it easier to jump to files if we're consistent with how they're named?

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+  "Resource" - {
+    "should encode resource id correctly" in {
+      Resource(
+        "arn:aws:s3:::my-bucket/*"
+      ).asJson.noSpaces shouldBe """"arn:aws:s3:::my-bucket/*""""
+    }
+  }
+
+  "Condition" - {
+    "should encode single condition correctly" in {
```
adamnfish, line 27: As above, I don't think this codec (or the principal one) are actually used as it stands. We could either drop them (and therefore also these tests), or change the `Statement` encoder to defer to other types for the encoding logic.

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+    "should encode statement with ID correctly" in {
+      val statementWithId = basicStatement.copy(id = Some("statement1"))
+      val expected = """{
+                         |"Effect":"Allow",
+                         |"Action":["iam:PutRolePolicy","iam:getRole"],
+                         |"Resource":["*"],
+                         |"Sid":"statement1"
+                         |}""".stripMargin.replaceAll("\\s", "")
+
+      statementWithId.asJson.noSpaces shouldBe expected
```
adamnfish, line 86: Would it help debugging test failures for all these JSON tests if we compare a pretty printed version? Something like `spaces2` in place of `noSpaces` would make it easier to visually parse the diffs that this test case genera, and would mean the expected value above could be formatted to match - without the need for the `replaceAll` call.

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
```diff
@@ -0,0 +1,269 @@
...
+        json should include(""""AWS":["AROAXXXXXXXXXXXXXXX"]""")
+      }
+
+      "with correct second principal" in {
+        val json = complexPolicy.asJson.noSpaces
+        json should include(""""Kubernetes":["ServiceAccount1"]""")
+      }
+
+      "with correct complete json string" in {
+        val json = complexPolicy.asJson.spaces2
```
adamnfish, line 231: Ah, like this! Do you think we should we do the same (`spaces2`) for the above?

app/aws/Federation.scala
```diff
@@ -141,24 +138,24 @@ object Federation {
...
+    val iamClient = IamClient.builder().credentialsProvider(provider).build()
 
     // remove access from assumed role
     val roleName = getRoleName(roleArn)
-    val getRoleRequest = new GetRoleRequest().withRoleName(roleName)
-    val role = Role(iamClient.getRole(getRoleRequest).getRole)
-    val roleRevocationPolicy =
-      RolePolicy(role, "janus-role-revocation-policy", revocationPolicyDocument)
-    // ^
-    // this name should match policy in cloudformation/federation.template.yaml
```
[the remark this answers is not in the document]
kelvin-chappell: Yes I'll put it back

configTools/src/main/scala/com/gu/janus/policy/Iam.scala
kelvin-chappell: I did it like this because the encoding spec appeared to show a single condition has a different encoding to multiple conditions but you're right it's only ever encoded as part of a policy so this is redundant and confusing.

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
kelvin-chappell: I'll include what this test was covering in the policy test and drop this.

configTools/src/main/scala/com/gu/janus/policy/Iam.scala
kelvin-chappell: Done in ab1e448ab285a1b31c1b9a57ccee15420e4f99ee

configTools/src/main/scala/com/gu/janus/policy/Iam.scala
kelvin-chappell: Removed in 7d52d1e0f6c82ee3037f26cf29b4ca4a883c8f95

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
kelvin-chappell: Renamed in b736c07f2cc05de8c671688b276e02f90f5e7b72

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
kelvin-chappell: Done in 7d52d1e0f6c82ee3037f26cf29b4ca4a883c8f95 and 9813a072e63b9bba5859c50ffe936b29ba674412

configTools/src/test/scala/com/gu/janus/policy/IamSpec.scala
kelvin-chappell: Done in 7108fccfe6e96ad3753ae781cc371b2f6e09f07c

configTools/src/test/scala/com/gu/janus/policy/IamTest.scala
```diff
@@ -0,0 +1,374 @@
...
+                       |        "203.0.113.1/24"
+                       |      ]
+                       |    }
+                       |  }
+                       |}""".stripMargin
+
+      statementWithCondition.asJson.spaces2 shouldBe expected
+    }
+
+    "should encode statement with multiple conditions correctly" in {
```
adamnfish, line 127: Do we need to test the case where we have multiple conditions of the same type, to check it merges properly?

```scala
conditions = List(
          Condition("aws:SourceIp", "StringEquals", List("203.0.113.0/24")),
          Condition(
            "aws:TokenIssueTime",
            "StringEquals", // <-- THE SAME HERE
            List("2025-02-27T09:33:01.000Z")
          )
        )
```

adamnfish (approved):
Let's do any more changes in follow-up PRs, because this has been open for a while! Apologies for the slow feedback loops for you.

LGTM, but let's test it carefully 👍

configTools/src/test/scala/com/gu/janus/policy/IamTest.scala
kelvin-chappell: Done in 7fbe8f459213feef1410a72b35feed80c4860b38

kelvin-chappell:
Smoke tests listed in PR description all passed in production.

PR #513, opened 2025-02-19
Set AWS STS client's region explicitly

kelvin-chappell:
The region for the AWS Java STS client was being read from the local users' config but this can't be relied on.
So now setting it explicitly in the code.


tjsilver (approved):
👍 This will make working locally much easier thank you!

PR #516, opened 2025-02-26
Refactor audit trail code

kelvin-chappell:
This change optimises and simplifies some of the code that manages interactions with the audit trail DB table.

## Changes
1. Removal of redundant [calls to describe table](https://github.com/guardian/janus-app/blob/f522bca83d9d90657634b038c7734a9871974161/app/aws/AuditTrailDB.scala#L22-L25).
   The name of the table and its key schema are already hardcoded throughout the code so it seems redundant to occasionally look these details up as well.
2. Removal of redundant table name parameter from various methods.
   As mentioned above, the table name is already hardcoded so need to pass it around as a parameter.
3. Replaced an awkward tuple with a case class, as suggested in [this comment](https://github.com/guardian/janus-app/pull/500#discussion_r1944762210).
4. Replaced repeated use of string literal table attribute names with shared constants.
   To make it clear where the same field is being used, what it is and avoid typos etc.
5. Some breakdown of functions to make them more readable.
6. Separation of concerns between [AuditTrail](https://github.com/guardian/janus-app/blob/f522bca83d9d90657634b038c7734a9871974161/app/logic/AuditTrail.scala) and [AuditTrailDB](https://github.com/guardian/janus-app/blob/f522bca83d9d90657634b038c7734a9871974161/app/aws/AuditTrailDB.scala) now clearer as the former does all marshalling of data between database and internal formats while AuditTableDB makes the requests.

## To test
These tests have been done locally and need to be repeated in production:

- [ ] Go to audit table and find a known access record => should appear correctly and correspond with what's in Dynamo DB table
- [ ] Sign into AWS console from Janus, go back to Janus and look at audit table => should see a new access record with fields as expected
- [ ] Request credentials from Janus and then look at audit table => should see a new access record with fields as expected

### DIFF
--- a/app/aws/AuditTrailDB.scala
+++ b/app/aws/AuditTrailDB.scala
@@ -30,7 +30,10 @@ object AuditTrailDB {
       .tableName(tableName)
       .keyConditions(
         Map(
-          attrEqualCondition(accountPartitionKeyName, AttributeValue.fromS(account)),
+          attrEqualCondition(
+            accountPartitionKeyName,
+            AttributeValue.fromS(account)
+          ),
           dateRangeCondition(startDate, endDate)
         ).asJava
       )
--- a/app/aws/AuditTrailDB.scala
+++ b/app/aws/AuditTrailDB.scala
@@ -30,7 +30,7 @@ object AuditTrailDB {
       .tableName(tableName)
       .keyConditions(
         Map(
-          attrCondition(accountPartitionKeyName, AttributeValue.fromS(account)),
+          attrEqualCondition(accountPartitionKeyName, AttributeValue.fromS(account)),
           dateRangeCondition(startDate, endDate)
         ).asJava
       )
@@ -50,7 +50,7 @@ object AuditTrailDB {
       .indexName(secondaryIndexName)
       .keyConditions(
         Map(
-          attrCondition(userNameAttrName, AttributeValue.fromS(username)),
+          attrEqualCondition(userNameAttrName, AttributeValue.fromS(username)),
           dateRangeCondition(startDate, endDate)
         ).asJava
       )
@@ -59,7 +59,7 @@ object AuditTrailDB {
     queryResult(dynamoDB, request)
   }
 
-  private def attrCondition(
+  private def attrEqualCondition(
       attrName: String,
       attrValue: AttributeValue
   ) =
--- a/app/aws/AuditTrailDB.scala
+++ b/app/aws/AuditTrailDB.scala
@@ -4,11 +4,7 @@ import com.gu.janus.model.AuditLog
 import logic.AuditTrail
 import org.joda.time.DateTime
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
-import software.amazon.awssdk.services.dynamodb.model.ComparisonOperator.{
-  BETWEEN,
-  EQ
-}
-import software.amazon.awssdk.services.dynamodb.model.KeyType.{HASH, RANGE}
+import software.amazon.awssdk.services.dynamodb.model.ComparisonOperator._
 import software.amazon.awssdk.services.dynamodb.model._
 
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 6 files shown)


app/logic/AuditTrail.scala
```diff
@@ -7,7 +7,71 @@ import org.joda.time.{DateTime, DateTimeZone, Duration}
...
+
+  val partitionKeyName = "j_account"
+  val sortKeyName = "j_timestamp"
+  val userNameAttrName = "j_username"
+  val durationAttrName = "j_duration"
+  val accessLevelAttrName = "j_accessLevel"
+  val accessTypeAttrName = "j_accessType"
+  val isExternalAttrName = "j_external"
+
+  private type DbAttr = (String, AttributeValue)
```
adamnfish, line 25: Would it simplify / clarify things to make this a case class instead of a type alias? Alternatively, is this bringing much value?

In the latter case, this type alias isn't used anywhere except right here (the next few lines), so maybe inling the real type would simplify the `AuditLogDbEntryAttrs` definition.

Considering the approach to instead use a case class, I notice that the tests for this code all look something like this: `attrs.partitionKey._2.s()`. Making this a case class with named fields would let us do something more descriptive than `_2`, something which I had to dig into while reviewing that part of this change.

I think either approahc to changing this would be fine. If you want to keep it as-is that's ok too, this isn't a big deal. It jumped out to me because I generally try to keep quite a high bar for things to qualify for type aliases. Type aliases bring the abstraction cost of having a new concept for a thing without the extra safety and ergonomics of a real typed wrapper with named fields.

app/aws/AuditTrailDB.scala
```diff
@@ -108,11 +59,21 @@ object AuditTrailDB {
     queryResult(dynamoDB, request)
   }
 
+  private def attrCondition(
```
adamnfish, line 62: I'm not 100% convinced this helper is worthwhile but if it is perhaps we could call it something like `attrEqualCondition`, to better convey that this is hard-coded to do an EQ check?

There are only two uses of this helper, and its definition isn't loads shorter than its invocation. Would it be simpler to write the builder out each time, especially since this would be consistent with the "BETWEEN" condition used just below for date range?

If IntelliJ did that underlined "Duplicate code fragment" thing then no worries, that's a clear sign wiser minds than I say this is definitely worthwhile 😁

app/controllers/Janus.scala
```diff
@@ -239,7 +238,7 @@ class Janus(
         janusData.access
       )
     } yield {
-      AuditTrailDB.insert(table, auditLog)
+      AuditTrailDB.insert(auditLog)
```
adamnfish, line 241: I realise this isn't part of your change, but while we're here anyway and tidying up the DB usage can we pull this up into the for comp steps? I don't like that we have an important side-effecting step in the yield block here (my apologies for the current state of things).

Something like `_ = AuditTrailDB.insert(auditLog)` as the last step would highlight that this is an important operation, separate from building our return value.

I'd do the same for the log line, but I don't feel as strongly we should do that now.

Equally, I'm very happy to tackle this along with a wider review of the codebase in a separate PR - I only mention it here because our diff has hit it.

app/aws/AuditTrailDB.scala
```diff
@@ -4,77 +4,33 @@ import com.gu.janus.model.AuditLog
...
-      .build()
+  def insert(auditLog: AuditLog)(implicit dynamoDB: DynamoDbClient): Unit = {
+    val auditLogDbAttrs = AuditLogDbEntryAttrs.fromAuditLog(auditLog)
+    val item = auditLogDbAttrs.toMap.asJava
+    val request =
+      PutItemRequest.builder().tableName(tableName).item(item).build()
     dynamoDB.putItem(request)
   }
 
-  private def toAttribValue(value: Any): AttributeValue = {
```
adamnfish, line 48: Delighted to see this go away, great change 👍

app/logic/AuditTrail.scala
```diff
@@ -7,7 +7,71 @@ import org.joda.time.{DateTime, DateTimeZone, Duration}
...
 
+import scala.util.Try
+
 object AuditTrail extends Logging {
+
+  val tableName = "AuditTrail"
+  val secondaryIndexName = "AuditTrailByUser"
+
+  val partitionKeyName = "j_account"
+  val sortKeyName = "j_timestamp"
```
adamnfish, line 18: There are a few places in the codebase after this change where we lose a bit of the connection to what's really going on, when we use this name.

From a DynamoDB implementation point of view, the DB's sort key is what we're using to implement date sorting/filtering. It makes sense to call this `sortKeyName` to highlight this important connection to the implementation detail. An example of this is https://github.com/guardian/janus-app/pull/516/files#diff-3fcccabe5ec8afa651c9696d905657fabb5b0ba56779b459c6150def64b600fbR76, makes great sense as-is.

However, it's also the audit log timestamp, and there are other use cases where it would sense for this name to be connected to our business domain. An example is the [date range condition's implementation above](https://github.com/guardian/janus-app/pull/516/files#diff-7a531d949edd241aaf1d3e5ff63af11e39624831e554bea05e04713b93b83022R76). In this case it's probably more helpful to know that we're using the timestamp here.

I don't have an answer for this, I think we're being pulled in both directions! One observation is that we refer to both domains [in this block](https://github.com/guardian/janus-app/pull/516/files#diff-8a966c9ceb1c443388c8049af7f39bdc17ea219bab1be4f0a9c924175d8613f2R55-R57), which I think helps with comprehension. The code right here tells us that this is the DB sort field and that it relates to the access time.

Maybe there's a name for this member that could convey that this is the timestamp *and* that it is the timestamp?

app/logic/AuditTrail.scala
```diff
@@ -7,7 +7,71 @@ import org.joda.time.{DateTime, DateTimeZone, Duration}
...
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue
 
+import scala.util.Try
+
 object AuditTrail extends Logging {
+
+  val tableName = "AuditTrail"
+  val secondaryIndexName = "AuditTrailByUser"
+
+  val partitionKeyName = "j_account"
```
adamnfish, line 17: This consolidation might also give us the opportunity to add a comment explaining the `j_` prefix, which is to avoid clashing with [DynamoDB reserved keywords](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ReservedWords.html). Up to you if you think this is relevant / useful?

adamnfish (approved):
Great stuff, thank you for doing the thinking to simplify this part of the codebase 👍 

I've added a few comments but nothing blocking, up to you if you're interested to tweak things a little more.

app/logic/AuditTrail.scala
kelvin-chappell: I'll have another look at this.  I initially modelled it as a case class but found that that made the instance creation code less readable.  I think I'll get rid of the type alias and improve the test code to make it clearer.

app/aws/AuditTrailDB.scala
kelvin-chappell: Yes this did come from an Intellij prompt!  I'll have another look at this one as well

app/logic/AuditTrail.scala
kelvin-chappell: Removed the type alias in e1c12578f05b9f2137f57c790e25f9a55574729b

app/aws/AuditTrailDB.scala
kelvin-chappell: Renamed in 21e2630e7fce7a6189807f151bbc061a39af9b2b

app/controllers/Janus.scala
kelvin-chappell: Done in d8d0f9e8c1a6ac653e3dd61d31cdabf48d4dc32c

app/logic/AuditTrail.scala
kelvin-chappell: Add some comments about the database attributes in e291ff3dba9068780ab0183d218b9dc3c578576e

kelvin-chappell:
Smoke tests all passing in production.

PR #522, opened 2025-03-03
Bump config-tools dependency in example project

kelvin-chappell:
This is a clean-up after #509 to ensure all known clients of the config-tools library are on the latest version.


adamnfish (commented):
Great stuff, thank you.

We should separately make sure this project builds (ideally as a separate job in CI that doesn't block merging). Great stuff remembering to update this here though 🙏

adamnfish (approved):

PR #524, opened 2025-03-05
Read Play secret from param store

kelvin-chappell:
This change integrates [Play secret rotation](https://github.com/guardian/play-secret-rotation) into the app.  The secret is now read from SSM param store, rather than from a config file.  We aren't yet automatically rotating the secret - that is still a manual step for now.

Once this has been fully set up, we will have an automatically updated secret, which will safe us some toil and keep us secure.

## Any additional notes?
Before this can be merged:
- [x] EC2 instance profile needs permission to read from parameter store

## To verify in production
- [ ] App loads successfully with no new errors in log

Co-authored-by: @tjsilver

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -39,10 +39,13 @@ class AppComponents(context: ApplicationLoader.Context)
     case _         => "DEV"
   }
 
+  // Reads Play secret from SSM
   val secretStateSupplier: SnapshotProvider =
     new parameterstore.SecretSupplier(
       TransitionTiming(
+        // When a new secret value is read it isn't used immediately, to keep all EC2 instances in sync.  The new value is used after the usageDelay has passed.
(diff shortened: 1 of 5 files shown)


app/AppComponents.scala
```diff
@@ -14,29 +16,49 @@ import router.Routes
...
+  // in that situation, it'll load assets directly from npm vs production, where they'll come from the bundled files
+  val mode: Mode = context.environment.mode
+
+  // Janus has no Code stage
+  private val stage = mode match {
+    case Mode.Prod => "PROD"
+    case _         => "DEV"
+  }
+
+  val secretStateSupplier: SnapshotProvider =
```
tjsilver, line 42: Could there be a comment about how the timing works?

tjsilver (approved):
👍

PR #525, opened 2025-03-05
Replace joda time with java time

kelvin-chappell:

[Joda time](https://github.com/JodaOrg/joda-time) was the standard time library in the Java ecosystem before Java 8 introduced its own [java.time module](https://dev.java/learn/date-time/).
This change updates time references to use the java.time module.

We have one fewer library and its dependencies to maintain.  
The AWS v2 clients use java.time Instants so we no longer need to convert between these and Joda time alternatives.

## Any additional notes?

Some of the Joda time types have been changed.

For timestamps, we now use [Instant](https://dev.java/learn/date-time/instant/) in place of [Joda DateTime](https://www.joda.org/joda-time/apidocs/org/joda/time/DateTime.html).
The Instant is a point in time that is a certain number of milliseconds since the modern epoch began.  So it is independent of timezone, which makes it easier to reason about.

For timespans, we now use [Duration](https://dev.java/learn/date-time/period-duration/) in place of [Joda Period](https://www.joda.org/joda-time/apidocs/org/joda/time/Period.html).
The Joda Period was a more generic concept.  In the java.time module, for simple timespans between nanoseconds and weeks there is a type Duration and for longer timespans where calendar support is required there is a type Period.  We use Duration because it covers the timespans we need and we don't need to consider calendar issues such as lengths of particular months, years etc.

## To test
Repeat these steps in production once deployed:

- [ ] Getting credentials for an account shows expiry at 19:00 local time
- [ ] Including an explicit session duration in request URL gives credentials that show expiry at expected local time
- [ ] Audit trail for a pre-existing session shows correct time start and duration
- [ ] Audit trail for a new session shows correct time start and duration
- [ ] Using date picker on audit trail page works correctly
- [ ] Revoking access to an account sets correct time threshold in new IAM condition
- [ ] Support dates show correctly


app/aws/Federation.scala
```diff
@@ -42,7 +42,7 @@ object Federation {
   def duration(
       permission: Permission,
       requestedSeconds: Option[Duration] = None,
-      timezone: Option[DateTimeZone] = None
+      clock: Option[Clock] = None
```
kelvin-chappell, line 45: Using a clock here instead of a timezone so that we can pass in either a real system clock representing a changing time in a particular timezone or a test fixed clock that has stopped at a given time.

app/controllers/Janus.scala
```diff
@@ -209,19 +210,23 @@ class Janus(
...
+        Instant.now(),
         janusData.access,
         janusData.admin,
         janusData.support
       )
-      duration = Federation.duration(permission, requestedDuration, tzOffset)
+      duration = Federation.duration(
+        permission,
+        requestedDuration,
+        tzOffset.map(Clock.system)
```
kelvin-chappell, line 228: This passes in the system clock set to the given timezone.

test/aws/FederationTest.scala
```diff
@@ -81,65 +80,72 @@ class FederationTest extends AnyFreeSpec with Matchers with JodaTimeUtils {
...
+          minute = 0,
+          zoneId = Some(ZoneOffset.ofHours(1))
+        ) { time =>
           duration(
             permission,
             None,
-            Some(DateTimeZone.forOffsetHours(1))
-          ) shouldEqual 3.hours
+            time
+          ) shouldEqual 4.hours
```
kelvin-chappell, line 148: The test expectation appeared to be wrong previously because the comparison is between 15:00 locally and 19:00 locally, which is 4 hours.  We appeared to be comparing local time (ie. UTC + 1 hour) with UTC.

app/aws/Federation.scala
```diff
@@ -42,7 +42,7 @@ object Federation {
   def duration(
       permission: Permission,
       requestedSeconds: Option[Duration] = None,
-      timezone: Option[DateTimeZone] = None
+      now: Option[ZonedDateTime] = None
```
kelvin-chappell, line 45: By passing in the current time, instead of time zone, we are able to test more easily because it's a static time.

tjsilver (approved):
Unable to test locally at the moment, but code looks good and tests seem very comprehensive and all pass. Great work!

test/aws/FederationTest.scala
adamnfish: Let's think about what we're trying to test, here? I think we're saying the system time should be irrelevant, and the test wants to check that the timezone the user provided is being used properly. I think this comes back to the added obscurity I mentioned above, mixing the current (server) time with the timezone logic. With the test as written I'd be expcting to see a timezone argument to the duration function, or at least see a timezone being applied to the current server time. If that doesn't make sense any more then I think this test (and its wording) should change?

test/testutils/TimeUtils.scala
```diff
@@ -0,0 +1,14 @@
+package testutils
+
+import java.time.{ZoneId, ZonedDateTime}
+
+trait TimeUtils {
+  def withSystemTime[T](hour: Int, minute: Int, zoneId: Option[ZoneId] = None)(
```
adamnfish, line 6: Be worth checkign there are any tests that need the timezone (especially after looking at the one I commented on above).

Arguably, if we want to make the server time independent of all the logic we shoulkd pick a random timezone and make sure that all the tests still pass!

adamnfish (changes requested):
Let's chat about the details on a few of these bits if this isn't clear? Apologies if I've misunderstood the code in my quick glance!

test/aws/FederationTest.scala
kelvin-chappell: Yes the way I read the test was that it's trying to verify there's nothing in the code that assumes a particular timezone.  But the way the test was previously it did seem to be comparing one timezone with another.

app/controllers/Janus.scala
```diff
@@ -55,7 +56,7 @@ class Janus(
   }
 
   def support = authAction { implicit request =>
-    val now = DateTime.now()
+    val now = Instant.now()
```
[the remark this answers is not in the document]
adamnfish: Occurs to me to mention that it doesn't necessarily need to be London here - it's good to use UTC everywhere on the server. Just that we need to make sure we're using 11am London time at that point (and doing that here is one way to keep the existing behaviour).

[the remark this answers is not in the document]
kelvin-chappell: I was confusing myself there.  Instants are implicitly in UTC.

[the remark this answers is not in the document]
kelvin-chappell: This is a very good point.  The support shift start time is set in the datafile, so it's essential that this arrives in the expected format.  I'll do some experimenting with the example subproject to make sure this works correctly: https://github.com/guardian/janus-app/blob/1b08633479199e86209abf06876cc632ad036441/example/src/main/scala/com/example/Support.scala#L43-L58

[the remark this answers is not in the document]
kelvin-chappell: Experimenting with the `example` subproject and with our internal Janus config this works as expected.  The date can be read correctly as either a Joda `DateTime` or a java.time `Instant`.  It's also read correctly as either 11 am UTC when London is in GMT or 10 am UTC when London is in BST.

app/aws/Federation.scala
```diff
@@ -38,32 +38,42 @@ object Federation {
...
+    *   or may not be honoured.
+    * @param optClock
+    *   [[Clock]] to use to determine the current time.
     */
   def duration(
       permission: Permission,
-      requestedSeconds: Option[Duration] = None,
-      timezone: Option[DateTimeZone] = None
+      requestedDuration: Option[Duration] = None,
+      optClock: Option[Clock] = None
```
kelvin-chappell, line 54: Now passing in an optional [Clock](https://dev.java/learn/date-time/clock/) instead so that the intention of the method still makes sense but we are also able to test it easily in different scenarios.  
A Clock represents a timezone but within that timezone the time can either be passing at a real-world rate or fixed.  A fixed clock always gives the same value for current time.  
According to the javadocs:
> Best practice for applications is to pass a Clock into any method that requires the current instant and time-zone.

We aren't doing this generally but it can be revisited elsewhere in a later PR.

test/aws/FederationTest.scala
kelvin-chappell: Federation.duration method signature changed so that it takes a Clock.  Also changed the test title.  And the test implementation now uses a property to test arbitrary timezones.

app/logic/Date.scala
```diff
@@ -1,132 +1,123 @@
...
-      Some {
-        simpleDateFormatter.parseDateTime(dateStr)
-      }
-    } catch {
-      case _: IllegalArgumentException => None
-    }
+  /** Parses a date string in the format "yyyy-MM-dd" and if it's valid returns
+    * the instant at the start of that date in UTC.
+    */
+  def parseDateStr(dateStr: String): Option[Instant] = {
```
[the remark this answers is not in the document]
kelvin-chappell: Renamed.

test/aws/AuditTrailDBTest.scala
```diff
@@ -2,27 +2,28 @@ package aws
...
     "insertion and querying" ignore {
-      val dateTime: DateTime =
-        new DateTime(2015, 11, 5, 17, 35, DateTimeZone.UTC)
+      val dateTime = ZonedDateTime.of(2015, 11, 5, 17, 35, 0, 0, UTC).toInstant
       val al = AuditLog(
         "account",
         "username",
         dateTime,
-        new Duration(3600000),
+        Duration.ofMillis(3600000),
```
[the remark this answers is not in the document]
kelvin-chappell: Now changed to `Duration.ofHours(1)`.

test/logic/AuditTrailTest.scala
```diff
@@ -93,19 +96,21 @@ class AuditTrailTest
         AuditTrail.auditLogFromAttrs(attrs).value should have(
           "account" as "account",
           "username" as "username",
-          "dateTime" as new DateTime(2015, 11, 4, 15, 22, DateTimeZone.UTC),
-          "duration" as new Duration(3600000),
+          "instant" as ZonedDateTime
+            .of(2015, 11, 4, 15, 22, 0, 0, UTC)
+            .toInstant,
+          "duration" as Duration.ofMillis(3600000),
```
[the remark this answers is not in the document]
kelvin-chappell: Done.

test/testutils/TimeUtils.scala
kelvin-chappell: We now have a test that checks the duration method works in any arbitrary timezone.

app/logic/Date.scala
```diff
@@ -1,132 +1,123 @@
...
-      .appendSeparator(", ")
-      .appendMinutes()
-      .appendSuffix(" minute", " minutes")
-      .appendSeparator(", ")
-      .appendSeconds()
-      .appendSuffix(" second", " seconds")
-      .toFormatter
-    minutesAndSeconds.print(period)
-  }
+  def formatDuration(duration: Duration): String = {
```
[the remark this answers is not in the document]
adamnfish: It doesn't have to go in this PR, but would something like this be clearer?

```scala
  def formatDuration(duration: Duration): String = {
    Seq(
      duration.toHours.toInt -> "hour",
      duration.toMinutesPart -> "minute",
      duration.toSecondsPart -> "second",
    ).collect {
      case (value, unit) if value == 1 => s"$value $unit"
      case (value, unit) if value > 0 => s"$value ${unit}s"
    }.mkString(", ")
  }
```

I'm not sure, whatever you think!

adamnfish (approved):
LGTM, thank you!

app/logic/Date.scala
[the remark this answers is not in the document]
kelvin-chappell: I like it! Updated in c8ce3885a55e3d193b049d6ebf310878c6c29f65

kelvin-chappell:
Smoke tests all passed in production.

PR #526, opened 2025-03-05
fix: console error in audit page

tjsilver:

There was a console error on the audit trail page about the submit functionality: `Uncaught TypeError: document.getElementById(...).on is not a function`.

The javascript code was calling an `.on('submit')` function for the button element. It turns out this functionality isn't required anyway, as the `<button>` element by default has the type of 'submit' and will submit the form without the need for javascript. 

I have removed the javascript and made the functionality of the button more explicit by specifying its parent form id and giving it the type of 'submit'.

No console error and datepicker continues to work as expected.


kelvin-chappell (approved):
🕵️ 👍

PR #527, opened 2025-03-05
refactor: update application.conf filters

tjsilver:

- Resolve the console errors observed in Chrome `Refused to load the font...` because it violates the following Content Security Policy directive: "font-src 'self' ` by adding `data:` to the font source CSP
- Remove the external fonts sources from the CSP - we now include fonts as part of the bundle. These sources are not referenced anywhere else in the codebase so should be safe to remove.

No console errors


kelvin-chappell (approved):
Looks good to me 👍

PR #533, opened 2025-03-20
feat: remove jquery dependency

tjsilver:

To remove the dependency on jQuery by converting all the jQuery code to plain Javascript.  

This will reduce the number of vulnerabilities on our packages. Currently there are 4 moderate-severity vulnerabilities on jQuery which will be eliminated with this change. 

## How to test locally - compare side-side-with PROD

- [x] Appearance of buttons: size at various breakpoints appears as expected
- [x] Appearance of buttons: when checkbox selected, buttons are greyed out as expected
- [x] Sidenav appears on mobile/tablet breakpoints as in PROD
- [x] Calendar picker on Audit Trail page: replacement native calendar picker functions as expected
- [x] Accordions on Support and Credentials pages function as expected
- [x] Can sign into AWS console
- [x] Can retrieve credentials and use them in command line locally
- [x] Can edit AWS profile name on Credentials page and credentials update correctly
- [x] Can revoke all credentials for an account and see policy condition has correct threshold time and console and local credentials unusable
- [x] Tested in Firefox
- [x] Tested Chrome
- [ ] Tested in Safari (will test on PROD)


adamnfish (approved):
LGTM 🥳

PR #535, opened 2025-03-21
Update the accounts information page

adamnfish:

This adds the account number to the accounts page, which is a critical piece of information for properly understanding which account is which.

This version of the page pulls the account number from the configured role, which is a nice simple way to go about it. A more complete solution would use the configured role to call `STS.getCallerIdentity` but this isn't currently required and would make this all more complicated.

We also tidy up and simplify the "owners" display. The previous implementation dates back to Janus' early versions, where there were only "devs" and "admins". This assumption no longer holds, so we instead list everyone that has any access, and tell users to look at the Janus access config to find out more detail.

Here again, a future version could show the JSON policy(s) available to each listed user, but since this duplicates the information in Janus' configuration there's no need to add this until it is asked for.

We have an open ticket to integrate an AWS account, it would be nice for us to be able to quickly and easily check whether the account is already integrated on the Accounts page (using its account ID).

This should also make the accounts page less confusing, since it doesn't include the old dev/admin assumption.

## Images

I'm ommitting screenshots here because they necessarily include account numbers and user details from the Janus configuration (and this is a public repository). I can screen share the layout, or you can run it locally to see how things look.

### DIFF
--- a/app/conf/Config.scala
+++ b/app/conf/Config.scala
@@ -20,7 +20,7 @@ object Config {
     requiredString(config, s"federation.$awsAccountAuthConfigKey.aws.roleArn")
 
   // extract aws account ID from Role ARN
-  private val AwsAccountId = """arn:aws:iam::(\d+):role/.*""".r
+  private val AwsAccountId = """arn:aws:iam::(\d+):role/.+""".r
   def accountNumber(
       awsAccountAuthConfigKey: String,
       config: Configuration
@@ -31,8 +31,9 @@ object Config {
         case AwsAccountId(accountId) => Success(accountId)
         case _ =>
           Failure(
-            new RuntimeException(
-              s"Could not extract account number from role ARN: $role"
+            new JanusConfigurationException(
+              s"Could not extract account number from role ARN $role",
+              s"federation.$awsAccountAuthConfigKey.aws.roleArn"
             )
           )
       }
@@ -102,8 +103,9 @@ object Config {
       val jsonCertStream =
         Try(new FileInputStream(serviceAccountCertPath))
           .getOrElse(
-            throw new RuntimeException(
-              s"Could not load service account JSON from $serviceAccountCertPath"
+            throw new JanusConfigurationException(
+              s"Could not load service account JSON",
+              serviceAccountCertPath
             )
           )
       ServiceAccountCredentials.fromStream(jsonCertStream)
@@ -118,7 +120,15 @@ object Config {
 
   private def requiredString(config: Configuration, key: String): String = {
     config.getOptional[String](key).getOrElse {
-      throw new RuntimeException(s"Missing required config property $key")
+      throw new JanusConfigurationException(
+        s"Missing required config property",
+        key
+      )
     }
   }
+
+  class JanusConfigurationException(message: String, location: String)
+      extends Throwable(
+        s"$message at $location"
+      )
 }
--- a/app/conf/Config.scala
+++ b/app/conf/Config.scala
@@ -13,11 +13,31 @@ import models._
 import play.api.Configuration
 
 import java.io.{File, FileInputStream}
-import scala.util.Try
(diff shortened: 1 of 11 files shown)


app/conf/Config.scala
```diff
@@ -13,12 +13,32 @@ import models._
...
 import java.io.{File, FileInputStream}
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 object Config {
   def roleArn(awsAccountAuthConfigKey: String, config: Configuration): String =
     requiredString(config, s"federation.$awsAccountAuthConfigKey.aws.roleArn")
 
+  // extract aws account ID from Role ARN
+  private val AwsAccountId = """arn:aws:iam::(\d+):role/.*""".r
```
kelvin-chappell, line 23: I suppose pedantically: `arn:aws:iam::(\d+):role/.+`

app/conf/Config.scala
```diff
@@ -13,12 +13,32 @@ import models._
...
+      awsAccountAuthConfigKey: String,
+      config: Configuration
+  ): Try[String] = {
+    for {
+      role <- Try(roleArn(awsAccountAuthConfigKey, config))
+      accountNumber <- role match {
+        case AwsAccountId(accountId) => Success(accountId)
+        case _ =>
+          Failure(
+            new RuntimeException(
```
kelvin-chappell, line 34: Do you think a more specific exception would be better?  Not an important point

app/controllers/Utility.scala
```diff
@@ -2,26 +2,44 @@ package controllers
...
+        )
+      }
+    // log any account number errors we accumulated
+    accountData
+      .collect { case (account, _, Failure(err)) =>
+        (account, err)
+      }
+      .foreach { case (account, err) =>
+        logger.warn(s"Couldn't lookup account number for ${account.name}", err)
+      }
```
kelvin-chappell, line 42: Perhaps this could be broken into subfunctions for clarity but comment makes it clear what's going on so again not important

app/views/accounts.scala.html
```diff
@@ -1,33 +1,49 @@
...
+                  Refer to <a href="@permissionsRepo">Janus' configuration</a> for full details on who has what access to each AWS account.
+              </p>
+          }
+          case None => {
+              <p>Refer to Janus' configuration for full details on who has what access to each AWS account.</p>
+          }
+        }
         <ul class="collapsible collapsible-accordion">
-        @for((account, owners) <- accountOwners) {
+        @for((account, owners, tryAccountNumber) <- accountOwners) {
```
kelvin-chappell, line 26: I suppose these aren't really account 'owner's
```suggestion
        @for((account, users, tryAccountNumber) <- accountOwners) {
```

kelvin-chappell (approved):
Looks good!  Tried it out locally and all working at different breakpoints for me.
I can't see anything wrong with the code.  I've made some very minor comments that can safely be ignored. 👍

app/views/accounts.scala.html
adamnfish: The old thinking behind the name "owners" here, is that people with access to an account are people that are respinsible for the account. In some sense, everyone that appears in this list is an "owner". I agree this becomes murkier now that the access levels are more tightly scoped in many cases, and I agree that the word owners isn't totally accurate here anymore. I decided to defer changing this for now, and I'll revisit it after we've done a bit of tidying, renaming and consolidating of the existing account access levels.

Great catch, I should have mentioned this in the PR description.

app/controllers/Utility.scala
adamnfish: I'm also always a bit nervous about having big controllers, particularly where it makes it hard to test the logic. This app doesn't use what i would now call the normal "for comp at the edge" style, but I agree with you this is still too far from the spirit of that!

I'll pull the logic pieces out into a testable locations so this Action is just a list of well-tested steps.

PR #541, opened 2025-03-28
Group GitHub actions version bumps

adamnfish:
We currently have multiple dependabot PRs open to bump version updates in GitHub actions. This takes time, because the app needs to be rebuilt after each PR is merged. For patch and minor version bumps on GitHub provided acitons, we're probably ok to tackle them as a group.


kelvin-chappell (approved):
👍

PR #549, opened 2025-04-07
feat: create account page with button to register a new passkey

tjsilver:

- Adds the functionality for a user to be able to register a passkey. This is part of our work to add a passkey authentication flow. 
	- The right-hand link (the user's name) has been changed from `/logout` to open a user account.
	- A new `/user-account` route has been added along with a landing page to support this.
	- The icon which appeared on the mobile view has been changed to a 'account box' icon, and this has also been added to the desktop view for consistency.
	- The wording of the link has changed to '[Firstname] [Lastname]'s account for clarity.
	- These changes currently only appear in DEV mode, as this feature is not ready to be released in PROD yet.

## Next steps

- Include the user's currently registered passkeys on the user's account landing page, populated from the database.

### DIFF
--- a/app/controllers/PasskeyController.scala
+++ b/app/controllers/PasskeyController.scala
@@ -12,12 +12,12 @@ import play.api.{Logging, Mode}
 
 /** Controller for handling passkey registration and authentication. */
 class PasskeyController(
-                         controllerComponents: ControllerComponents,
-                         authAction: AuthAction[AnyContent],
-                         host: String,
-                         janusData: JanusData
-                       )(implicit mode: Mode, assetsFinder: AssetsFinder)
-  extends AbstractController(controllerComponents)
+    controllerComponents: ControllerComponents,
+    authAction: AuthAction[AnyContent],
+    host: String,
+    janusData: JanusData
+)(implicit mode: Mode, assetsFinder: AssetsFinder)
+    extends AbstractController(controllerComponents)
     with Logging {
 
   private val appName = "Janus"
--- a/app/controllers/PasskeyController.scala
+++ b/app/controllers/PasskeyController.scala
@@ -95,4 +95,11 @@ class PasskeyController(
 
     result.merge
   }
+
+  def showUserAccountPage: Action[AnyContent] = authAction {
+    implicit request =>
+      Ok(views.html.userAccount(request.user, janusData))
+  }
 }
+
+
--- a/app/controllers/PasskeyController.scala
+++ b/app/controllers/PasskeyController.scala
@@ -1,0 +1,98 @@
+package controllers
+
+import com.webauthn4j.data._
+import com.webauthn4j.data.attestation.statement.COSEAlgorithmIdentifier
+import com.webauthn4j.data.client.challenge.{Challenge, DefaultChallenge}
+import play.api.libs.json.{JsNumber, JsString, Json, Writes}
+import play.api.mvc._
+
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.Base64
+import scala.jdk.CollectionConverters._
+
+class PasskeyController(controllerComponents: ControllerComponents)
+    extends AbstractController(controllerComponents) {
+
+  implicit val publicKeyCredentialTypeWrites: Writes[PublicKeyCredentialType] =
+    Writes(`type` => JsString(`type`.getValue))
+
+  implicit val coseAlgorithmIdentifierWrites: Writes[COSEAlgorithmIdentifier] =
+    Writes(alg => JsNumber(alg.getValue()))
+
+  implicit val publicKeyCredentialParametersWrites
+      : Writes[PublicKeyCredentialParameters] =
+    Writes(params =>
+      Json.obj(
+        "type" -> params.getType,
+        "alg" -> params.getAlg
+      )
+    )
+
+  implicit val publicKeyCredentialRpEntityWrites
+      : Writes[PublicKeyCredentialRpEntity] =
+    Writes(rp =>
+      Json.obj(
+        "id" -> rp.getId,
+        "name" -> rp.getName
+      )
+    )
+
+  implicit val publicKeyCredentialUserEntityWrites
+      : Writes[PublicKeyCredentialUserEntity] =
+    Writes(user =>
+      Json.obj(
+        "id" -> Base64.getUrlEncoder.withoutPadding
+          .encodeToString(user.getId),
+        "name" -> user.getName,
+        "displayName" -> user.getDisplayName
+      )
+    )
+
+  implicit val challengeWrites: Writes[Challenge] =
+    Writes(challenge =>
+      JsString(
+        Base64.getUrlEncoder.withoutPadding
+          .encodeToString(challenge.getValue)
+      )
+    )
+
+  implicit val publicKeyCredentialCreationOptionsWrites
+      : Writes[PublicKeyCredentialCreationOptions] =
+    Writes(options =>
+      Json.obj(
+        "challenge" -> options.getChallenge,
+        "rp" -> options.getRp,
+        "user" -> options.getUser,
+        "pubKeyCredParams" -> options.getPubKeyCredParams.asScala
+      )
+    )
+
+  def registrationOptions: Action[AnyContent] = Action { _ =>
+    val appDomain = "example.com"
+    val appName = "Janus"
+    val userId = "user-id"
+    val userName = "user-name"
+    val userDisplayName = "user-display-name"
+    val alg = COSEAlgorithmIdentifier.ES256
+    val rp = new PublicKeyCredentialRpEntity(appDomain, appName)
+    val user = new PublicKeyCredentialUserEntity(
+      userId.getBytes(UTF_8),
+      userName,
+      userDisplayName
+    )
+    val challenge = new DefaultChallenge()
+    val pubKeyCredParams = List(
+      new PublicKeyCredentialParameters(
+        PublicKeyCredentialType.PUBLIC_KEY,
+        alg
+      )
+    ).asJava
+    val options = new PublicKeyCredentialCreationOptions(
+      rp,
+      user,
+      challenge,
+      pubKeyCredParams
+    )
+    Ok(Json.toJson(options))
+  }
+}
--- a/app/controllers/PasskeyController.scala
+++ b/app/controllers/PasskeyController.scala
@@ -1,132 +1,98 @@
 package controllers
 
-import com.webauthn4j.WebAuthnManager
-import com.webauthn4j.credential.CredentialRecord
-import com.webauthn4j.data._
-import com.webauthn4j.data.attestation.statement.COSEAlgorithmIdentifier
-import com.webauthn4j.data.client.Origin
-import com.webauthn4j.data.client.challenge.{Challenge, DefaultChallenge}
-import com.webauthn4j.server.ServerProperty
-import play.api.libs.json.{JsNumber, JsString, Json, Writes}
+import com.gu.googleauth.AuthAction
+import com.gu.janus.model.JanusData
+import com.webauthn4j.util.Base64UrlUtil
+import io.circe.Json
+import io.circe.syntax.EncoderOps
+import logic.Passkey
+import models.Passkey._
 import play.api.mvc._
+import play.api.{Logging, Mode}
 
-import java.nio.charset.StandardCharsets.UTF_8
-import java.util.Base64
-import scala.jdk.CollectionConverters._
+/** Controller for handling passkey registration and authentication. */
+class PasskeyController(
+                         controllerComponents: ControllerComponents,
+                         authAction: AuthAction[AnyContent],
+                         host: String,
+                         janusData: JanusData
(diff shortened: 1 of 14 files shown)


frontend/passkeys.js
```diff
@@ -0,0 +1,35 @@
...
+        });
+    }
+
+    registerPasskeyButtons.forEach(function(btn) {
+        btn.addEventListener('click', function(e) {
+            e.preventDefault();
+            registerPasskey().then(function() {
+                console.log('Registration successful');
+            }).catch(function(err) {
+                console.error('Error during registration:', err);
```
kelvin-chappell, line 30: I wonder whether this logging adds any value to just looking at the network requests.  I think we also saw that a 403 was appearing as a success so it might even be misleading.

frontend/passkeys.js
```diff
@@ -0,0 +1,35 @@
+document.addEventListener('DOMContentLoaded', function() {
+    'use strict';
+
+    const registerPasskeyButtons = document.querySelectorAll('#register-passkey');
```
kelvin-chappell, line 4:
```suggestion
    const registerPasskeyButton = document.querySelectorAll('#register-passkey');
```

frontend/passkeys.js
```diff
@@ -0,0 +1,35 @@
...
+
+
+    async function registerPasskey() {
+        const response = await fetch('/passkey/registration-options') //fetch PublicKeyCredentialCreationOptions as JSON string
+        const publicKeyCredentialCreationOptionsJSON = await response.json() // convert to JSONObject
+        const credentialCreationOptions = PublicKeyCredential.parseCreationOptionsFromJSON(publicKeyCredentialCreationOptionsJSON); // convert to PublicKeyCredentialCreationOptions
+        const publicKeyCredential = await navigator.credentials.create({publicKey: credentialCreationOptions}); // create PublicKeyCredential
+        console.log('publicKeyCredential: ', publicKeyCredential);
+        const registrationResponseJSON = publicKeyCredential.toJSON(); // convert to JSONObject
```
kelvin-chappell, line 13: Not sure that the comments here are useful.  They were originally included as a guide because this whole block of code was copied from https://webauthn4j.github.io/webauthn4j/en/#generating-a-webauthn-credential-key-pair

frontend/passkeys.js
```diff
@@ -0,0 +1,35 @@
...
+
+    const registerPasskeyButtons = document.querySelectorAll('#register-passkey');
+
+
+    async function registerPasskey() {
+        const response = await fetch('/passkey/registration-options') //fetch PublicKeyCredentialCreationOptions as JSON string
+        const publicKeyCredentialCreationOptionsJSON = await response.json() // convert to JSONObject
+        const credentialCreationOptions = PublicKeyCredential.parseCreationOptionsFromJSON(publicKeyCredentialCreationOptionsJSON); // convert to PublicKeyCredentialCreationOptions
+        const publicKeyCredential = await navigator.credentials.create({publicKey: credentialCreationOptions}); // create PublicKeyCredential
+        console.log('publicKeyCredential: ', publicKeyCredential);
```
kelvin-chappell, line 12: I don't think there's any need to log anything here unless you think it's useful.  If the public key isn't created properly or is invalid the `register` call will fail anyway

frontend/passkeys.js
```diff
@@ -0,0 +1,35 @@
...
+        const registrationResponseJSON = publicKeyCredential.toJSON(); // convert to JSONObject
+        await fetch('/passkey/register', {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'text/plain',
+                'Csrf-Token': `${registerPasskeyButtons[0].getAttribute('csrf-token')}`
```
kelvin-chappell, line 18: I wonder if it would be more elegant to pass the value of csrf-token into `registerPasskey` from the event listener.  We know that there will be a `registerPasskeyButtons[0]` at this point because otherwise `registerPasskey` wouldn't be called but it still seems a bit odd.

app/views/passkeyRegistration.scala.html
```diff
@@ -0,0 +1,14 @@
...
+@(user: UserIdentity, janusData: JanusData)(implicit request: RequestHeader, mode: Mode, assetsFinder: AssetsFinder)
+
+@main("Registration", Some(user), janusData) {
+
+    <div class="container">
+
+        <h1 class="header orange-text">Registration Mash</h1>
+
+    </div>
+}
```
kelvin-chappell, line 14: I think this page is redundant now

app/controllers/PasskeyController.scala
```diff
@@ -31,9 +31,14 @@ class PasskeyController(
...
+  def showRegistrationPage: Action[AnyContent] = authAction {
+    implicit request =>
+      Ok(views.html.passkeyRegistration(request.user, janusData))
+  }
+
   private def apiResponse[A](
-      action: => Try[A]
-  )(implicit encoder: Encoder[A]): Result =
+                              action: => Try[A]
+                            )(implicit encoder: Encoder[A]): Result =
```
kelvin-chappell, line 41: Formatting has gone a bit awry here.
Running `sbt scalafmtAll` will sort it out.

app/controllers/PasskeyController.scala
```diff
@@ -31,9 +31,14 @@ class PasskeyController(
     case Mode.Prod => "Janus-Prod"
   }
 
+  def showRegistrationPage: Action[AnyContent] = authAction {
+    implicit request =>
+      Ok(views.html.passkeyRegistration(request.user, janusData))
+  }
```
kelvin-chappell, line 37: It looks as though this action is redundant now as the view isn't used, unless I'm missing something.

conf/routes
```diff
@@ -29,6 +29,9 @@ GET     /logout                     controllers.AuthController.logout
 
 GET     /passkey/registration-options   controllers.PasskeyController.registrationOptions
 POST    /passkey/register               controllers.PasskeyController.register
+GET         /passkey/register                    controllers.PasskeyController.showRegistrationPage
```
kelvin-chappell, line 32: Appears to be redundant, if page is no longer needed:
```suggestion
```

frontend/passkeys.js
tjsilver: I did get an error during registration when working locally, so I think it might be useful to keep it. It highlighted a type error.

frontend/passkeys.js
tjsilver: I have refactored this so use a 'querySelector' method and optional chaining.

app/views/main.scala.html
```diff
@@ -36,7 +36,11 @@
                             <li><a href="/support">Support</a></li>
                             @if(hasAccess(username(user), janusData.admin)) { <li><a href="/superuser">Admin</a></li> }
                             <li><a href="/revoke">Revoke</a></li>
-                            <li><a href="/logout">@user.firstName @user.lastName</a></li>
+                            @if(mode == Mode.Dev) {
+                                <li><a href="/user-account">@user.firstName @user.lastName's account <i class="material-icons right">account_box</i></a></li>
```
adamnfish, line 40: Can we leave this as-is at the moment, ideally not even linking to it in the menu? I see this feature as being exploratory at the moment. It isn't going to be turned on for people so there's no need to put the work-in-progress in front of them

We can leave the new endpoint (as we have for registration/authentication), and just have it be there only for people that already knwo what they're looking for.

When the time does come to link to this, I'm not sure the `'s account` suffix adds much? Maybe no harm adding the icon, but I think it's fine to leave the name as-is.

An alternative (and what I'd aim for in the future) would be to pull the avatar URL from the UserIdentity that comes back from Google. This field is an option, so we'd have a fallback user image as well in the cases where we don't get the avatar coming through. In my testing, the avatar comes through just fine for me.

Two caveats to this approach are:
1. We'd need to add the domain for the avatar images, which is a domain that contains Google's user-uploaded content
2. This domain contains Google's user-uploaded content, so one might make the case that we don't want to allow it on a secure system like Janus

Something to think about later, if you agree!

frontend/passkeys.js
```diff
@@ -0,0 +1,31 @@
...
+            },
+            body: JSON.stringify(registrationResponseJSON)
+        });
+    }
+
+    registerPasskeyButton?.addEventListener('click', function(e) {
+            e.preventDefault();
+            const csrfToken = this.getAttribute('csrf-token');
+            registerPasskey(csrfToken).catch(function(err) {
+                console.error(err);
```
adamnfish, line 27: Fine for now, but it'd be good to add some more robust error handling to help improve the user experience. In particular, our API endpoints return a nice JSON objec tthat contains a friendly user message, with the goal being that this is something we can display.

Again though, I think this is something to add later 👍 
today we're just putting these pages in to get an end-to-end test of the process so this looks ok to me.

app/views/main.scala.html
```diff
@@ -88,6 +96,12 @@ <h5 class="white-text">Useful links</h5>
             <script src="@assetsFinder.path("frontend/janus.js")"></script>
         }
 
+        @if(mode == Mode.Dev) {
```
adamnfish, line 99: Would it be easier to import the passkeys file into janus.js so it's a single JS bundle? If we're going to unconditionally include both scripts anyway it might be simpler and more performant to let Parcel bundle it all up together? I like that this is separate, though!

No harm structurally for now since we're still in exploration-mode, but in the future I think I wouljd aim to separate the passkey logic from the Janus DOM integrations. The logic for registering and authenticating (probably inluding the API urls and our API response handling) would make a nice "module" for passkey support, which could then be used by a new section in `janus.js`, which hooks this behaviour onto any page(s) where it is required. This approach would guide us towards leaving `janus.js` as our "entrypoint" and havign passkey.js be a separate file (module) that janus.js imports and uses.

app/controllers/PasskeyController.scala
```diff
@@ -101,4 +101,8 @@ class PasskeyController(
         )
       )
       .toTry
+
+  def showUserAccountPage: Action[AnyContent] = authAction { implicit request =>
```
adamnfish, line 105: I agree with this, perfect place to put it as a placeholder. If we turn this into a proper `me` page in the future it'll might go in the Utilities controller, and could include useful personal features like upcoming support slots and a link to (and/or records from) the user's personal audit trail history.

app/views/main.scala.html
adamnfish: Ah I've spotted that this is only in dev mode! Let's keep dev and prod consistent if we can. Maybe if we want this to be behind a flag we could add an application configuration flag to enable / disable all these features, and that way we'd be able to quickly and easy test it in either mode locally (and enable / disable it in PROD).

I reckon this is something for later though, maybe just ahead of conditionally adding this behaviour to the console/credentials endpoints.

adamnfish (changes requested):
I've blocked it for now to make sure we chat about the nav change! Otherwise I'm mostly sharing thoughts for the future 👍

app/views/main.scala.html
tjsilver: I agree. Have done this in https://github.com/guardian/janus-app/pull/549/commits/655ed2c8107cdecd3642d45a0bbd43b954ed3faf.

app/views/main.scala.html
tjsilver: Done in [0bbd209](https://github.com/guardian/janus-app/pull/549/commits/0bbd2098e6c0e6b595ba514f291ada3322257850).

app/views/main.scala.html
```diff
@@ -45,7 +49,11 @@
                             @if(isSupportUser(username(user), Instant.now(), janusData.support)) { <li><a href="/support">Support</a></li> }
                             @if(hasAccess(username(user), janusData.admin)) { <li><a href="/superuser">Admin</a></li> }
                             <li><a href="/revoke">Revoke</a></li>
-                            <li><a href="/logout">@user.firstName @user.lastName <i class="material-icons right">power_settings_new</i></a></li>
+                            @if(mode == Mode.Dev) {
```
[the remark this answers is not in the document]
tjsilver: Done in [4ee5fd4](https://github.com/guardian/janus-app/pull/549/commits/4ee5fd408b14134e27e66aa7ffb5064e715f3fd6).

app/views/userAccount.scala.html
```diff
@@ -0,0 +1,28 @@
...
+            <div class="col s12 m12 l9">    
+                <div class="card-panel">          
+                    <p>You can have up to two passkeys to log in to Janus. One should be on-device (e.g. Touch ID) and the other off-device (e.g. security key or authenticator app).
+                    </p>
+                    <div><button id="register-passkey" class="btn" csrf-token="@{CSRF.getToken.get.value}">Register a new passkey</button></div>
+                </div>
+            </div>
+            <div class="col s12 m12 l3">
+                <div class="card-panel">
+                        <p>You are logged in as @user.firstName @user.lastName.</p><div><a href="/logout"><button class="btn">Log out</button></a></div>
```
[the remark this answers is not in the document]
tjsilver: Done in [b10ede3](https://github.com/guardian/janus-app/pull/549/commits/b10ede3fdccc54ebe7d6dddda9d7b3a2b816d39c).

adamnfish (approved):

PR #552, opened 2025-04-11
feat: add endpoints for passkey authentication

kelvin-chappell:
This change adds two new endpoints to support authentication of endpoints by passkey:
* GET /passkey/auth-options (to initiate an auth flow)
* POST /passkey/authenticate (to verify the authentication attached to the request)

This should unblock the clientside development to do a full passkey lifecycle.

## Any additional notes?
The authenticate endpoint will be replaced by a custom action that can be attached to any POST route in the next PR.

## Testing locally
There is a test harness included in 7b8335bfdae9732f289ace8d7d37dcfbcfc7044b, which allows the new endpoints to be tested (if you checkout that commit).  You can then test the whole passkey lifecycle with the temporary `/passkey/register` (registration) and `/passkey/authenticate` (authentication) endpoints in a browser.  
The test harness has been removed from the final PR. 

## Related
* #542 
* #549


app/logic/Passkey.scala
```diff
@@ -97,6 +98,36 @@ object Passkey {
...
+    Try {
+      new PublicKeyCredentialRequestOptions(
+        challenge,
+        null,
+        null,
+        null,
+        null,
+        null
+      )
+    }.recoverWith(exception =>
```
adamnfish, line 119: I think all these `recoverWith` calls would be simpler with `.adaptError` from Cats, but we can very easily retrofit that approach.

app/logic/Passkey.scala
```diff
@@ -154,12 +185,104 @@ object Passkey {
...
+    *
+    * @param jsonResponse
+    *   Json response from browser containing authentication data.
+    */
+  def parsedAuthentication(
+      jsonResponse: String
+  ): Try[AuthenticationData] =
+    Try {
+      webAuthnManager.parseAuthenticationResponseJSON(jsonResponse)
+    } recoverWith {
```
adamnfish, line 211: This syntax isn't encouraged in modern Scala, although it does work. [The style guide](https://docs.scala-lang.org/style/method-invocation.html#higher-order-functions) recommends using `.` for method calls.

app/aws/PasskeyDB.scala
```diff
@@ -54,16 +77,136 @@ object PasskeyDB {
...
         )
       )
     )
+  }
+
+  def extractCredential(
+      response: GetItemResponse,
+      user: UserIdentity
+  ): Try[CredentialRecord] = {
+    if (response.hasItem) {
```
adamnfish, line 124: Not super urgent, but I think I'd look at whether this if statement (and the resulting duplication of error handling at the bottom as well as the nesting it causes) could be inlined. Maybe making this a for comp and doing `item <- response.item().toRight(<err>).toTry` as the first line would make this simpler? Perhaps something to fiddle with when we do another pass on this.

adamnfish (approved):
Looks ok to me! A couple of little comments, but as we've discussed I think it's ok to merge this without feeling like we need to get it perfect up front while we explore the feature.

This is with the caveat that I haven't properly studied the passkey logic, here as well this will be easier for us to do a full review of when we have the full end-to-end flow in place to tinker with.

app/logic/Passkey.scala
```diff
@@ -154,12 +185,104 @@ object Passkey {
...
+            engineerMessage =
+              s"Authentication parsing failed: ${err.getMessage}",
+            httpCode = BAD_REQUEST,
+            causedBy = Some(err)
+          )
+        )
+      case err =>
+        Failure(
+          JanusException(
+            userMessage = "Bad authentication object",
```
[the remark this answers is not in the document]
kelvin-chappell: I think it's useful to keep them distinct but I agree that the user messages could be a bit more helpful

app/logic/Passkey.scala
```diff
@@ -97,6 +98,36 @@ object Passkey {
...
+    * [[https://webauthn4j.github.io/webauthn4j/en/#generating-a-webauthn-assertion]].
+    */
+  def authenticationOptions(
+      user: UserIdentity,
+      challenge: Challenge = new DefaultChallenge()
+  ): Try[PublicKeyCredentialRequestOptions] =
+    Try {
+      new PublicKeyCredentialRequestOptions(
+        challenge,
+        null,
```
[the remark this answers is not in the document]
kelvin-chappell: 😕 I'm not sure I understand.  This method isn't doing much.  It just creates a random string and returns it to the caller wrapped in a `PublicKeyCredentialRequestOptions` instance.  So in fact the `Try` will only fail if the challenge passed into it is `null`.  The `Try` block is really a complete overkill here.  But not sure what would be timing out

app/logic/Passkey.scala
kelvin-chappell: Interesting!  There are lots of cats features that I need to learn about.  I'll look at that in a follow-up PR

app/aws/PasskeyDB.scala
kelvin-chappell: Thanks!  I'll keep a note of this and follow up

app/logic/Passkey.scala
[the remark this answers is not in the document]
kelvin-chappell: Will save this for a follow-up PR.

app/aws/PasskeyDB.scala
```diff
@@ -54,16 +77,136 @@ object PasskeyDB {
...
+            s"Credential data not found for user ${user.username}: GetItem response: $response",
+          httpCode = INTERNAL_SERVER_ERROR,
+          causedBy = None
+        )
+      )
+    }
+  }
+
+  /** See https://www.w3.org/TR/webauthn-1/#sign-counter */
+  def updateCounter(user: UserIdentity, authData: AuthenticationData)(implicit
```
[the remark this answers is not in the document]
kelvin-chappell: Following our discussion I've added a bit more to the method scaladoc in 95f87ddc5089445ea4a60c16750c006645cb4780

app/logic/Passkey.scala
```diff
@@ -97,6 +98,36 @@ object Passkey {
...
+        null,
+        null,
+        null,
+        null
+      )
+    }.recoverWith(exception =>
+      Failure(
+        JanusException(
+          userMessage = "Failed to create registration options",
```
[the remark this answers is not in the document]
kelvin-chappell: Fixed in 5b365b856acc4ebaa9bf6748cab77f239bee00d8

app/logic/Passkey.scala
kelvin-chappell: Fixed in a26c9d6699bdad934a5460eacedb300ff2ab78be

kelvin-chappell:
Making a call on the `/passkey/auth-options` in production is giving the correct response.

PR #553, opened 2025-04-11
Add a play run hook that starts the DynamoDB container

adamnfish:
Adds a play run hook that starts the DynamoDB via `docker compose`. This automates an essential manual step in getting this application running locally.

This saves a step that is a bit tedious, and that is hard to discover for users coming to janus-app for the first time.


tjsilver (approved):
This is really useful. I've tested locally and observed the database start and successfully queried it.

PR #555, opened 2025-04-16
feat: enable https on the frontend app when running locally

tjsilver:

To enable https on the frontend app when running locally. 

Also removes the 'hot module reloading' option from parcel as this feature is currently not working. We can revisit this in the future.

We are able to run the app locally in Safari as well as Firefox and Chrome.

There is an associated PR in the [`janus` repo](https://github.com/guardian/janus/pull/4728) to create the nginx mapping.


kelvin-chappell (approved):
👍

adamnfish (approved):

PR #558, opened 2025-04-24
feat: add eslint for frontend project

tjsilver:

- Adds linting to the frontend javascript files using eslint and our config. 
- Fixes the linting errors found.

We are applying best practices to our frontend code by ensuring the files are linted.

- [ ] Tested locally


kelvin-chappell (approved):
👍

PR #563, opened 2025-05-02
fix: update roleSessionName for account revocation

kelvin-chappell:
The configuration of privileges is slightly different in Prod and Dev environments.  This changes the username that can revoke access to accounts to reflect this difference.

We will be able to test revocation of all current access to an AWS account locally in a dev environment.

## Any additional notes?
See guardian/janus#4737 for configuration of Guardian's resources for local testing.


tjsilver (approved):
👍

kelvin-chappell:
Revocation still working in Dev and in Prod since deploy.

PR #565, opened 2025-05-06
refactor: migrate from Circe to Play Json for passkey data encoding

kelvin-chappell:
We're using Circe for Json handling of the passkey responses largely because we were using it in the configTools subproject and then over-generalised.  It's easier to use Play-json in a Play app as it slightly simplifies the handling of json request bodies and gives us consistent handling of Json throughout the app.

As we're working in the Play framework it makes sense to use its features unless there's a reason not to.


kelvin-chappell:
Can still register a passkey successfully in Prod after deploy.

PR #573, opened 2025-05-14
feat: Add lastUsedTime to Passkey model

kelvin-chappell:
This adds a timestamp for the last time the passkey was used to authenticate an action to the database.
We then show the timestamp on the passkey section of the user account page.

Can be used for troubleshooting and to give user confidence.

## Images
Screenshot 2025-05-14 at 13 19 55


tjsilver (approved):
I tested this locally and it works as expected in Chrome. However, in Firefox, this fails, perhaps because of the attestation being None leading to the passkey not being recognised as existing in the database.

kelvin-chappell:
Tested in Prod and all working well.

PR #571, opened 2025-05-13
feat: add delete passkey functionality

tjsilver:
## What is the purpose of this change?

This pull request introduces functionality for deleting passkeys, including backend, frontend, and UI changes. The most important changes include adding new methods for deleting passkeys in the database, creating a corresponding API endpoint, updating the UI to include delete buttons, and implementing frontend logic to handle passkey deletion.

### Backend Changes:
* **Database Methods**: Added `loadCredentialByName` and `deleteById` methods to the `PasskeyDB` object to query and delete passkeys from the DynamoDB table. These methods handle exceptions and provide meaningful error messages.
* **Controller Endpoint**: Introduced a `deletePasskey` method in `PasskeyController` to handle HTTP requests for deleting passkeys. It validates the request, deletes the passkey using `PasskeyDB`, and redirects the user upon success.
* **Routes**: Added a new route (`POST /passkey/delete`) to map the delete passkey API endpoint to the `deletePasskey` controller method.

### Frontend Changes:
* **UI Updates**: Updated the delete button in the `userAccount.scala.html` file to include `data-passkey-id` and `data-passkey-name` attributes for easier identification of the passkey being deleted.
* **JavaScript Logic**: 
  - Added `deletePasskey` and `setUpDeletePasskeyButtons` functions in `passkeys.js` to handle the deletion process, including confirmation dialogs and form submission.
  - Integrated the new delete button setup in `janus.js` to initialize event listeners on page load. [[1]](diffhunk://#diff-157d328996ad7679d704789ae8346a8f08d8aa777aab1f412acc445879b352f8R63-R109) [[2]](diffhunk://#diff-88cc93f042de9edc2d7dcae4d80e117f44c66f39596c94ed1fe7810574937367R299)
## Testing

To test this locally, run this branch and try to create and then delete a passkey.

PR #578, opened 2025-05-20
feat: Improve UX of user account page

tjsilver:
### Before - no passkeys

image

## Before - one or more passkeys

image

### After - no passkeys

image

### One or more passkeys

image

## Any additional notes?

I have reverted the two-passkey limit as this was not required.

I have also changed this PR remove the error messaging and fetch api use from this PR as I will use Play session flash messages rather than client-side toast messages in a separate PR.

### DIFF
--- a/app/views/userAccount.scala.html
+++ b/app/views/userAccount.scala.html
@@ -17,7 +17,7 @@
         <h3 class="header orange-text">Passkeys</h3>
 
         <div class="row">
-            <div class="col s12 m12 l9">
+            <div class="col s12">
                 <div class="card-panel">
                     <p>You can have up to two passkeys to log in to Janus. One should be on-device (e.g. Touch ID) and the other off-device (e.g. security key or authenticator app).
                     </p>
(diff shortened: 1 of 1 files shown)


kelvin-chappell (commented):
This is a big improvement!  I just made some nitpicking comments.

app/views/userAccount.scala.html
```diff
@@ -12,47 +12,62 @@
...
-        <div class="row">
-            <div class="col s12 m12 l9">
-                <div class="card-panel">
-                    <p>You can have up to two passkeys to log in to Janus. One should be on-device (e.g. Touch ID) and the other off-device (e.g. security key or authenticator app).
-                    </p>
-                    <div><button id="register-passkey" class="btn" csrf-token="@{CSRF.getToken.get.value}">Register a new passkey</button></div>
-                </div>
-            </div>
-        </div>
+        <p>Passkeys are a secure way to log in without a password. They are stored on your device and are used to authenticate you when you access Janus credentials or the AWS console.</p>
```
kelvin-chappell, line 19: They can also be off-device

kelvin-chappell:
> Now that we are using the fetch API we can make the app more RESTful by changing the passkey `delete` endpoint to a `DELETE` from a `POST` as suggested in a [previous PR](https://github.com/guardian/janus-app/pull/571) by @kelvin-chappell . We can do this in a separate PR.

💯

frontend/passkeys.js
```diff
@@ -86,18 +155,20 @@ export function setUpDeletePasskeyButtons(selector) {
...
                 console.error('No passkey ID found');
-                M.toast({ html: 'Error: Passkey ID not found', classes: 'rounded red' });
+                M.toast({ html: DOMPurify.sanitize('Error: Passkey ID not found'), classes: 'rounded red' });
                 return;
             }
             
             if (!csrfToken) {
                 console.error('No CSRF token found');
-                M.toast({ html: 'Error: Security token not found', classes: 'rounded red' });
+                M.toast({ html: DOMPurify.sanitize('Error: Security token not found'), classes: 'rounded red' });
```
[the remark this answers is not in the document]
tjsilver: I asked Copilot about the sanitising of the hard-coded messages and its response was that although it's not required now, it could be in the future if we replace the messages to include variables, so it's safer to sanitise all the messages regardless... WDYT @kelvin-chappell ?

app/views/userAccount.scala.html
tjsilver: fixed in https://github.com/guardian/janus-app/pull/578/commits/7c16bcff38e5ac19ed81f02c2adb09259516c10f

tjsilver:
> > Now that we are using the fetch API we can make the app more RESTful by changing the passkey `delete` endpoint to a `DELETE` from a `POST` as suggested in a [previous PR](https://github.com/guardian/janus-app/pull/571) by @kelvin-chappell . We can do this in a separate PR.
> 
> 💯

Not true anymore - see PR description.

kelvin-chappell (approved):
👍

PR #579, opened 2025-05-20
refactor: Use webauthn library native JSON encodings

kelvin-chappell:
Instead of using handwritten `play-json` encodings this change makes use of the Json annotations that have already been attached to the models of the `webauthn4j` library.

This is to avoid having to remember to update Json encodings whenever we use an extra field.  It also gives us more consistency in that we're now using the `webauthn4j` library to encode and to decode Json in web requests and responses.

## Any additional notes?
This is to pave the way for adding `excludeCredentials` argument to our `PublicKeyCredentialCreationOptions`, which in turn will stop the possibility of registering duplicate and therefore redundant passkeys.

### DIFF
--- a/app/controllers/PasskeyController.scala
+++ b/app/controllers/PasskeyController.scala
@@ -8,14 +8,13 @@ import com.gu.janus.model.JanusData
 import logic.AccountOrdering.orderedAccountAccess
 import logic.UserAccess.{userAccess, username}
 import logic.{Date, Favourites, Passkey}
-import models.JanusException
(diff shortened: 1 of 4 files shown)


tjsilver (approved):
Looks good! Will this also help when we come to try and identify the passkey type further down the line?

kelvin-chappell:
> Will this also help when we come to try and identify the passkey type further down the line?

Yes, we should be able to change `val attestation: AttestationConveyancePreference = NONE` to get more info

PR #582, opened 2025-05-21
feat: Improve passkey registration to prevent duplicate entries

kelvin-chappell:
We now send an [excludeCredentials](https://web.dev/articles/webauthn-exclude-credentials) attribute in our pre-registration options call, which contains details of the user's existing passkeys.  If they try to register a passkey of the same type twice they will get an `InvalidStateError`, which we trap with a warning toast.

This will ensure that we don't end up with redundant and useless passkeys in the Janus database.  If a passkey of the same type is registered twice, the first one will be overwritten on the authenticator and therefore its record in the Janus database will be useless.  By intercepting the registration on the browser we avoid a new passkey of the same type being generated at all.

### DIFF
--- a/frontend/passkeys.js
+++ b/frontend/passkeys.js
@@ -20,14 +20,19 @@ export async function registerPasskey(csrfToken) {
             passkeyName: passkeyName
         });
     } catch (err) {
-        console.error('Error during passkey registration:', err);
-        M.toast({ html: 'Failed to register passkey. Please try again.', classes: 'rounded red' });
+        if (err.name === 'InvalidStateError') {
+            console.warn('Passkey already registered:', err);
+            M.toast({html: 'This passkey has already been registered.', classes: 'rounded orange'});
+        } else {
+            console.error('Error during passkey registration:', err);
+            M.toast({html: 'Failed to register passkey. Please try again.', classes: 'rounded red'});
+        }
     }
 }
 
 export function setUpRegisterPasskeyButton(selector) {
     const registerButton = document.querySelector(selector);
-    if (!registerButton) { return };
+    if (!registerButton) { return }
 
     registerButton?.addEventListener('click', function (e) {
         e.preventDefault();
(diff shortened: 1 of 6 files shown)


tjsilver (commented):
I tested this locally and noticed that the warning message was only displayed in Chrome, and in Firefox the 'failed to register' error message was displayed instead. Is this expected?

kelvin-chappell:
> I tested this locally and noticed that the warning message was only displayed in Chrome, and in Firefox the 'failed to register' error message was displayed instead. Is this expected?

No, this is my limited testing.  Thanks for pointing it out!  I'll investigate.

frontend/passkeys.js
```diff
@@ -20,14 +20,19 @@ export async function registerPasskey(csrfToken) {
...
         });
     } catch (err) {
-        console.error('Error during passkey registration:', err);
-        M.toast({ html: 'Failed to register passkey. Please try again.', classes: 'rounded red' });
+        if (err.name === 'InvalidStateError') {
+            console.warn('Passkey already registered:', err);
+            M.toast({html: 'This passkey has already been registered.', classes: 'rounded orange'});
+        } else {
+            console.error('Error during passkey registration:', err);
+            M.toast({html: 'Failed to register passkey. This may be because this passkey has already been registered or it could be a transient issue, in which case please try again.', classes: 'rounded red'});
```
tjsilver, line 28: The message is very long and it might not stay up long enough to read it all. We could increase the display time and/or shorten the message to something like this?
```suggestion
            M.toast({html: 'Registration failed: this passkey may have already been registered, or it could be a transient issue in which case please try again.', classes: 'rounded red'});
```

tjsilver (approved):
Apart from a nit about the message length, this looks good!

frontend/passkeys.js
kelvin-chappell: When we include the possibility that the registration has been cancelled we'll have to start using footnotes!

PR #583, opened 2025-05-21
fix (passkeys): Add custom encoding for PublicKeyCredentialUserEntity

kelvin-chappell:
This change ensures that the user ID is base64url encoded when it's presented to [PublicKeyCredential.parseCreationOptionsFromJSON](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/parseCreationOptionsFromJSON_static) in the browser.

Without this change, certain users whose names are encoded in a way that doesn't produce a base64url encoded string will be rejected in some browsers, at least Chrome.
We see error:
> Error during passkey registration: Error during passkey registration: EncodingError: Failed to execute 'parseCreationOptionsFromJSON' on 'PublicKeyCredential': 'user.id' contains invalid base64url data


tjsilver (approved):
Tested locally and successfully registered a passkey on Chrome! 👍

PR #588, opened 2025-05-28
feat: enhance passkey name validation to prevent duplicates

tjsilver:

This pull request enhances the passkey management functionality by introducing duplicate name validation for passkeys and improving the structure of the HTML table to facilitate this validation. The most important changes include adding a data attribute to the passkey name table cells, implementing logic to fetch existing passkey names, and validating against duplicates during user input.

### Enhancements to passkey management:

* [`app/views/userAccount.scala.html`](diffhunk://#diff-374d8384fc5549c89757384e7a301bd56c2701178a0ecda3457140b0b40cd9b0L40-R40): Added `data-passkey-name="true"` attribute to passkey name table cells to enable easier identification of passkey names in the DOM.

* [`frontend/passkeys.js`](diffhunk://#diff-157d328996ad7679d704789ae8346a8f08d8aa777aab1f412acc445879b352f8R234-R249): Added logic to fetch existing passkey names from the HTML table and store them in a case-insensitive list for validation. This includes handling potential errors during DOM parsing.

### Validation improvements:

* [`frontend/passkeys.js`](diffhunk://#diff-157d328996ad7679d704789ae8346a8f08d8aa777aab1f412acc445879b352f8R269-R277): Updated the passkey name validation logic to check for duplicate names (case-insensitive) and display an appropriate error message if a duplicate is detected.

* [`frontend/passkeys.js`](diffhunk://#diff-157d328996ad7679d704789ae8346a8f08d8aa777aab1f412acc445879b352f8L266-R302): Refined the validation flow to ensure error messages are displayed consistently for invalid input, including duplicate names, excessive length, and disallowed characters.

### DIFF
--- a/app/aws/PasskeyDB.scala
+++ b/app/aws/PasskeyDB.scala
@@ -283,4 +283,27 @@ object PasskeyDB {
   }.recoverWith(err =>
     Failure(JanusException.failedToDeleteDbItem(user, tableName, err))
   )
+
+  /** Checks if a passkey name already exists for the given user. */
+  def passkeyNameExists(
+      user: UserIdentity,
+      passkeyName: String
+  )(implicit dynamoDB: DynamoDbClient): Try[Boolean] = Try {
+    val expressionValues = Map(
+      ":username" -> AttributeValue.fromS(user.username),
+      ":passkeyName" -> AttributeValue.fromS(passkeyName)
+    )
+    val request = QueryRequest
+      .builder()
+      .tableName(tableName)
+      .keyConditionExpression("username = :username")
+      .filterExpression("passkeyName = :passkeyName")
+      .expressionAttributeValues(expressionValues.asJava)
+      .build()
+
+    val response = dynamoDB.query(request)
+    response.hasItems && !response.items().isEmpty
+  }.recoverWith(err =>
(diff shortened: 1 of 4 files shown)


frontend/passkeys.js
```diff
@@ -231,6 +231,22 @@ function getPasskeyNameFromUser() {
         const alphanumericRegex = /^[a-zA-Z0-9 _-]*$/;
         const maxLength = 50; // Maximum character limit
 
+        // Get existing passkey names
+        let existingPasskeyNames = [];
+        try {
+            const passkeysTable = document.querySelector('.card-panel table');
```
kelvin-chappell, line 237: Is this a reliable way to find the table?  Perhaps it should have an ID

kelvin-chappell (commented):
As we're going to have to do similar validation on the server-side would it be simpler just to do it there?  Otherwise we have to keep the rules in sync on the client-side and the server-side

tjsilver:
> As we're going to have to do similar validation on the server-side would it be simpler just to do it there? Otherwise we have to keep the rules in sync on the client-side and the server-side

I think my idea was to do this client side so that we don't have to do it server-side. But if we think we have to do this server-side for other reasons than UX, then I agree, let's do it there.

PR #590, opened 2025-05-29
feat: Enable bypassing of passkey authentication for testing

kelvin-chappell:
This change makes the bypass of passkeys consistent with the use of passkeys so that protected links are now POSTs instead of GETs in both cases.

This fixes an error introduced in #585 where, if passkeys were being bypassed, the link would do a GET but there was no corresponding GET route available on the server.

## Any additional notes?
We can now do the same thing with the live links so that they can be tested with and without passkey authentication.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,3 +140,4 @@ crashlytics-build.properties
 .metals/
 .vscode/
 .bloop/
+/project/metals.sbt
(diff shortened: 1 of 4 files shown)


tjsilver (approved):
Tested locally and when passkeys were bypassed (i.e. I didn't add the cookie), instead of an error I got straight to the aws creds/console, and with the cookie, I got the passkey prompt and flow. Looks good! 👍

PR #591, opened 2025-05-30
fix: Improve passkey authentication error handling

kelvin-chappell:
Currently when a user tries to authenticate with a passkey that they haven't registered, they get a 500 response rather than the expected 401 response.  This happened because the failure manifested itself as a failure to find a corresponding passkey in the database.  This change fixes that.

The user gets the correct response and has a better understanding of what happened.

## Any additional notes?
There will be a follow-up PR coming that will stop the authentication flow before an authentication attempt is made when the user has no passkeys registered.  But this PR will still be a useful fallback in the case that a direct authentication attempt is made.


tjsilver (approved):
👍 Tested locally and worked as expected

PR #594, opened 2025-05-30
chore: Upgrade to Scala 3

kelvin-chappell:
This is the minimum change required to move to Scala 3.
This unlocks many improvements that can be made incrementally to take advantage of Scala 3 features.

The `janus-config-tools` artefact has been set to cross-compile so that it will continue to work in Scala 2.13 codebases.
This has been tested by making [a preview release](https://github.com/guardian/janus-app/releases/tag/v5.0.0-PREVIEW.kcscala3.2025-05-30T1454.3e777165) and successfully generating config in janus [guData](https://github.com/guardian/janus/blob/d6d771726da32e44c02f7816d4d9f6015a060a44/guData/src/main/scala/com/gu/janus/data/Main.scala) with it.

It's best to hide whitespace changes when reviewing.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -139,9 +139,19 @@ lazy val configTools = (project in file("configTools"))
   .enablePlugins(SbtTwirl)
   .settings(
     commonSettings,
-    crossScalaVersions := Seq("2.13.16", scalaVersion.value),
-    // No previous Scala 3 version yet
-    // TODO: remove this when we have released a Scala 3 version
+    crossScalaVersions := Seq(
+      "2.13.16",
+      scalaVersion.value
+    ),
+    /*
+     * The release process tries to find the previously released version
+     * so that it can check for the level of compatibility
+     * to determine the version number of the next release.
+     * If there is no previous release, the process breaks.
+     * To get around this we temporarily use the latest 2.13 release make the comparison.
+     *
+     * TODO: remove this when we have released a Scala 3 version
+     */
     mimaPreviousArtifacts := Set(
       organization.value %% name.value % "4.0.0" cross CrossVersion.for3Use2_13
     ),


build.sbt
```diff
@@ -20,14 +18,18 @@ val commonDependencies = Seq(
...
-  scalaVersion := "2.13.16",
-  scalacOptions ++= Seq(
-    "-deprecation",
-    "-feature",
-    "-unchecked",
-    "-release:11",
-    "-Xfatal-warnings"
-  ),
+  scalaVersion := "3.3.6",
+  scalacOptions ++= {
```
adamnfish, line 22: Something like this avoids having to cross reference the common options on each version path:

```scala
scalacOptions ++= Seq(
  "-feature",
  "-release:11"
) ++ (CrossVersion.partialVersion(scalaVersion.value) match {
  case Some((2, 13)) => Seq("-Xfatal-warnings")
  case Some((3, _))  => Seq("-Werror")
  case _             => Seq.empty
}),
```

build.sbt
```diff
@@ -135,10 +136,13 @@ lazy val root: Project = (project in file("."))
 lazy val configTools = (project in file("configTools"))
   .enablePlugins(SbtTwirl)
   .settings(
-    addCompilerPlugin(
-      "org.typelevel" %% "kind-projector" % "0.13.3" cross CrossVersion.full
-    ),
     commonSettings,
+    crossScalaVersions := Seq("2.13.16", scalaVersion.value),
+    // No previous Scala 3 version yet
```
adamnfish, line 141: It took me a while to figure this out, maybe these comments could be a little more explicit?

Instead of `No previous Scala 3 version yet` perhaps we could say what we're doing here - ensuring binary compatibility during the migration. For the TODO below, in the absence of an existing Scala 3 release to compare against automatically, we're explicitly pointing it at the previous release (`v4.0.0`)?

See what you think though!

adamnfish (approved):
A couple of minor comments, but LGTM.

build.sbt
kelvin-chappell: I agree - it wasn't very clear.  I hope I've improved it in 810cd39edd6fc815f2cbb6d0fbbd9d04b1576a2b

build.sbt
```diff
@@ -140,10 +138,23 @@ lazy val root: Project = (project in file("."))
...
+     * so that it can check for the level of compatibility
+     * to determine the version number of the next release.
+     * If there is no previous release, the process breaks.
+     * To get around this we temporarily use the latest 2.13 release make the comparison.
+     *
+     * TODO: remove this when we have released a Scala 3 version
+     */
+    mimaPreviousArtifacts := Set(
+      organization.value %% name.value % "4.0.0" cross CrossVersion.for3Use2_13
+    ),
```
rtyley, line 157: Ah, this is an interesting approach! 

This problem - of the MIMA compatibility check failing when new artifacts are introduced (because there are no previous artifacts to compare against for compatibility testing) is a known and annoying issue:

* https://github.com/guardian/gha-scala-library-release-workflow/issues/33

...in the past, I've dealt with this simply by _disabling_ the automated compatibility testing when releasing new artifacts. Again, it needs to be reinstated after the new artifact is released.

PR #602, opened 2025-06-06
fix: Update CSP to allow inline data images

kelvin-chappell:
We show inline Base64 PNG images against passkeys.  This change updates the CSP to allow these to be shown.


tjsilver (approved):
👍

PR #603, opened 2025-06-06
chore: Update deprecated Play config

kelvin-chappell:
This change updates some config settings that have been deprecated in recent releases of Play.
* Switched to using the [CSP Filter](https://www.playframework.com/documentation/3.0.x/CspFilter) instead of the deprecated CSP features of the SecurityHeadersFilter.
* langs config key change

Will stop broken functionality in future upgrades of Play.

This is a no-op change.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   release:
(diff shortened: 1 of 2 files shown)


tjsilver (approved):
👍

PR #604, opened 2025-06-06
chore: Cross-compile the config-tools project

kelvin-chappell:
Compile config-tools to Scala 2.13 and in Scala 3.

Now that we are publishing two versions of the config-tools library, we need to make sure that the project is always able to compile to both.


tjsilver (approved):
👍

PR #609, opened 2025-06-09
feat(passkeys): Discourage discoverable passkeys

kelvin-chappell:
This pull request updates the `Passkey` config in `app/logic/Passkey.scala` to modify the handling of resident (discoverable) keys during authentication.  It now 'discourages' resident keys.

Passkeys can't be discovered at authentication time unless they're referenced in the pre-authentication options call, which keeps the choice of passkeys under the control of the Janus server.

## Any additional notes?
See https://www.corbado.com/glossary/resident-key#resident-key-vs-non-resident-keys-whats-the-difference
Resident keys are designed to make the flow smoother when the user is unknown.
In our case, there's always an identifiable signed-in user and we know their credential IDs so we don't need to be able to 'discover' passkeys.


tjsilver (commented):
Tested locally and was no longer able to register a security key - it wasn't given as an option in either Firefox or Chrome. Is this the desired behaviour?

PR #610, opened 2025-06-10
chore: Update deprecated Play config (Attempt 2)

kelvin-chappell:
This is another attempt at #603 - to update config to improve content security policy (CSP) and resolve deprecated config warnings.
On that occasion it gave us error:

> Refused to load the script ... because it violates the following Content Security Policy directive: "script-src 'nonce-...' 'unsafe-inline' 'unsafe-eval' 'strict-dynamic' https: http:"

The difference in this PR is that nonce attributes have been added to HTML script elements to enable scripts to be loaded in browser with reduced risk of XSS injection, following the [Play CSP filter documentation](https://www.playframework.com/documentation/3.0.x/CspFilter#Using-CSP-in-Page-Templates).

Will stop broken functionality in future upgrades of Play.

This should be a no-op change.

## Any additional notes?
Reproduced problem seen in Production by selectively applying same CSP locally.

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -13,6 +13,7 @@ import play.api.mvc.{ActionBuilder, AnyContent, EssentialFilter}
 import play.api.routing.Router
 import play.api.{ApplicationLoader, BuiltInComponentsFromContext, Logging, Mode}
 import play.filters.HttpFiltersComponents
+import play.filters.csp.CSPComponents
 import router.Routes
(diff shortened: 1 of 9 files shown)


tjsilver (approved):
🤞

kelvin-chappell:
Now working in Production and no console errors, all functionality appears to be working and no reports of anything broken.
The logs show deprecation warnings have gone.
Dev config file has been updated.

PR #611, opened 2025-06-10
fix: Ensure wallclock option is removed only if it exists

tjsilver:

There was a bug where the wallclock element was null if the time was before 9am. This was probably introduced when the Janus frontend project was converted from jQuery to Javascript.

This remedies the bug by adding a null check for the element.


kelvin-chappell (approved):
👍

PR #612, opened 2025-06-10
feat(passkeys): Support mock multiSelect component

kelvin-chappell:
This adds passkey protection to the multi-account selection button on the mock home page.
The mock multiSelect component is a copy of the real one but with attributes added to the element with class `multiple-credentials__link`.

Brings passkey protection to all the sensitive points of the app.
Tested locally.

## Any additional notes?
The first commit in this PR fixes a bad label on an input field that Chrome browser reported.


tjsilver (approved):
Tested locally in Chrome, Firefox and Safari(!) and works as expected 👍

PR #613, opened 2025-06-10
fix(passkeys): Allow all authenticator types

kelvin-chappell:
By setting `AuthenticatorAttachment = PLATFORM` we were inadvertently restricting registerable passkeys to device-only.  ie. the setting is a requirement setting rather than a preference setting.
In this change, we allow any authenticator type, whether platform, cross-platform or hybrid.

This means that users have maximum choice of passkey.  In particular, we can register Yubikeys as passkeys.


tjsilver (approved):
Tested locally and was able to register a security key in either Chrome or Firefox 👍

PR #618, opened 2025-06-11
fix(passkeys): Handle null authenticatorAttachment

kelvin-chappell:

This pull request introduces a temporary workaround in the `registerPasskey` function to address browser compatibility issues with the `authenticatorAttachment` field when its value is `null`. The change ensures consistent behavior across browsers while maintaining the intended functionality.

Browser compatibility adjustment:

* `frontend/passkeys.js`: Added logic to remove the `authenticatorAttachment` field from `authenticatorSelection` if its value is `null`, addressing issues where Safari disallows `null` and Firefox issues warnings, while ensuring no functional impact.

The registration flow will run in any browser without warnings.

### DIFF
--- a/app/controllers/PasskeyController.scala
+++ b/app/controllers/PasskeyController.scala
@@ -2,11 +2,12 @@ package controllers
 
 import aws.PasskeyChallengeDB.UserChallenge
 import aws.{PasskeyChallengeDB, PasskeyDB}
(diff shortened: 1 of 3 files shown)


tjsilver (approved):
Tested before and after locally in Safari and can confirm that the error has gone, was able to register a passkey 👍

PR #624, opened 2025-06-17
feat(passkeys): Redirect to user account page if no passkeys

kelvin-chappell:

Currently, if a user tries to access a protected link when they haven't registered a passkey they are left on the page with a 400 error.
In this PR, the user is redirected to the user-account page where they are encouraged to register a passkey.

This improves the UX by making it clear what's happening and nudging the user to take the right action.

### User account page for user with no passkeys:
image


tjsilver (approved):
Tested locally and worked as described on Chrome and Firefox. 👍

PR #625, opened 2025-06-18
ci: Add CI workflow for Example app

kelvin-chappell:
It provides CI for the Example app included in this repo.  Currently there's no CI because it's a standalone app with its own SBT project.

To feel confident that the Example app is working.


tjsilver (approved):
👍

PR #627, opened 2025-06-19
feat: improve passkey frontend security and refactor to organise code better

tjsilver:


app/views/userAccount.scala.html
```diff
@@ -73,19 +72,17 @@ <h5>No Passkeys Registered</h5>
...
+                        <button id="register-passkey" class="btn">
                             Register a new passkey
                         </button>
                     </div>
                 </div>
             </div>
         </div>
     </div>
-    <div id="passkey-name-modal" class="modal">
+    <div id="passkey-name-modal" class="modal" role="dialog" aria-modal="true" aria-labelledby="modal_label" inert>
```
tjsilver, line 83: See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert for more info about what the `inert` attribute does.

frontend/passkeys.js
```diff
@@ -43,31 +44,14 @@ export async function registerPasskey(csrfToken) {
     } catch (err) {
         if (err.name === 'InvalidStateError') {
             console.warn('Passkey already registered:', err);
-            M.toast({html: 'This passkey has already been registered.', classes: 'rounded orange'});
+            displayToast('This passkey has already been registered.', messageType.warning);
         } else {
             console.error('Error during passkey registration:', err);
-            M.toast({html: 'Registration failed: this passkey may have already been registered, or it could be a transient issue in which case please try again.', classes: 'rounded red'});
+            displayToast('Registration failed: passkey may already be registered', messageType.error);
```
tjsilver, line 50: Tried to shorten this message here, WDYT @kelvin-chappell ?

frontend/passkeys.js
```diff
@@ -43,31 +44,14 @@ export async function registerPasskey(csrfToken) {
     } catch (err) {
         if (err.name === 'InvalidStateError') {
             console.warn('Passkey already registered:', err);
-            M.toast({html: 'This passkey has already been registered.', classes: 'rounded orange'});
```
tjsilver, line 46: Should this be an error rather than a warning?

frontend/passkeyNameModal.js
```diff
@@ -0,0 +1,164 @@
...
+
+        const handleKeyPress = (e) => {
+            if (e.key === 'Enter') {
+                e.preventDefault();
+                submitButton.click();
+            }
+        };
+
+        // Add event listeners using the named functions
+        input.addEventListener('input', handleInput);
```
adamnfish, line 137: My "old person" spidey sense tells me its best to wait until submission before applying validation results to the DOM like this. The reason is that it is distracting for users to be told things are wrong while they're still typing, especially if it's something that they know will be fixed by the time they've finished.

The normal flow would be to validate on submission, and clear validation errors as soon as the user starts editing the field. This model gets out the way while the user is working, trusts them to do the right thing, and then points out problems when they're back in "reading the screen" mode.

Of course, in this model we'd be intercepting the submission if validation errors exist!

A middle ground is to validate "on blur", but this is typcially only valid for forms with multiple fields (which this isn't) and still leads to reduced speed of entry when users stop to correct errors between each field.

This isn't a huge deal for a single field like this one, and the distraction is mitigated by the fact that in practice (today) the types of validation we're doing won't kick in immediately.

Up to you!

frontend/utils/toastMessages.js
```diff
@@ -0,0 +1,43 @@
...
+            html: DOMPurify.sanitize(message),
+            classes: `rounded ${type}`,
+    });
+}
+
+/**
+ * Displays flash messages from the server as toasts
+ * @param {Object} flashMessage Object containing flash messages by type
+ */
+export function displayFlashMessages(flashMessage) {
```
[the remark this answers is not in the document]
tjsilver: No, we don't need to consider multiple messages of one type as the code only sets one flash message per redirect, and form validation (`Validation.formattedErrors()`)  handles multiple errors by concatenating multiple errors into a single string.

flashMessages pluralised here: [7485a80](https://github.com/guardian/janus-app/pull/627/commits/7485a80b8a95ddb1e89fecdd3136737052a46b6d)

frontend/passkeyNameModal.js
tjsilver: I have added a debounce to the real-time validation to reduce excessive validation calls here: 482be19

frontend/passkeyNameModal.js
```diff
@@ -0,0 +1,257 @@
+import DOMPurify from 'dompurify';
+import M from 'materialize-css';
+import { displayToast, messageType } from './utils/toastMessages';
+
```
kelvin-chappell, line 4: Perhaps file should have header so that casual browser can see what it's for.

frontend/utils/toastMessages.js
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+export const messageType = {
+    error: 'red',
+    warning: 'orange',
+    info: 'blue',
+    success: 'green'
+} 
+
```
kelvin-chappell, line 12: Is spacing around this file deliberate?

frontend/passkeyNameModal.js
tjsilver: Will do this in the next PR too! This will be refactored into a `modalUtils` file.

tjsilver: The next PR is https://github.com/guardian/janus-app/pull/635

PR #629, opened 2025-06-23
refactor: Introduce PasskeyFlow enum

kelvin-chappell:
This change introduces a new `PasskeyFlow` enum, which is to enable us to distinguish between passkey challenges that have been stored for an authentication flow and those that have been stored for a registration flow.

This will ensure that the right challenge is compared during the passkey verification process.
There are various scenarios where challenges can get confused.  But this becomes more of an issue when passkeys are used to authenticate passkey registration requests.

## Any additional notes?
The change adds a new `flow` key attribute to items in the `PasskeyChallenge` Dynamo table.
The table configuration will need to be updated to reflect this change.


tjsilver:
> > Looks good, but testing locally failed even with destroying/recreating the passkeys table. Is this because the table configuration has not also been updated?
> 
> Yes, you have to destroy/create the PasskeyChallenges table. I'll add that to the PR description

Of course! 🙄

PR #632, opened 2025-06-25
chore: Give Scala Steward access to Example project

kelvin-chappell:
This repo contains a separate SBT project holding an example app.  This change ensures that Scala Steward keeps the dependencies of the project up to date along with the main SBT projects that are configured at the repo root level.

All Scala dependencies kept up to date throughout repo.


tjsilver (approved):
👍

PR #634, opened 2025-06-25
chore: Use scalafmt in Example project

kelvin-chappell:
The CI workflow introduced in #625 assumed that scalafmt was in use but it wasn't so the [latest run failed](https://github.com/guardian/janus-app/actions/runs/15874679007/job/44759334899#step:4:22).

Integrating scalafmt caused the reformatting of one file.

CI will pass and code will have consistent formatting.


tjsilver (approved):
👍

PR #645, opened 2025-07-07
fix: Stop sending null content-type request headers

kelvin-chappell:
We were seeing warnings in the server log:
`Illegal header: Illegal 'content-type' header: Invalid input 'EOI', expected tchar, OWS or '/'`.

This is because `auth-options` requests were being sent from passkeys.js with a `null` content type header.
The root cause is that `fetchWithCsrf` is being called with `null` as the contentType argument.

The optional arguments of `fetchWithCsrf` are `contentType` and `body` so in this PR the argument list has been rearranged to put these at the end.

This makes calls clearer and avoids inadvertently sending `null` as the `content type` header.

### DIFF
--- a/frontend/passkeys.js
+++ b/frontend/passkeys.js
@@ -8,9 +8,9 @@ const passkeyApi = {
   async fetchWithCsrf(
     url,
     method = "POST",
-    body = null,
-    contentType = "application/x-www-form-urlencoded",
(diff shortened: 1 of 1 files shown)


tjsilver (approved):
Tested locally and worked as expected. 👍

PR #666, opened 2025-08-14
chore: Review dependencies before merging

kelvin-chappell:
This change will warn us in advance if we are potentially introducing a vulnerability to the main branch.

The workflow will fail if a high level vulnerability is being introduced.

## Any additional notes?
Running as a separate workflow because it only makes sense to run when there are changes to a pull request.

If successful, could auto-add to other repos.

See ['Dependency Review' comment](https://github.com/guardian/janus-app/pull/666#issuecomment-3187517197) below for output.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -15,13 +15,13 @@ jobs:
       pull-requests: write
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
       # Required because Dependabot Alerts doesn't natively support Scala
-      - uses: guardian/setup-scala@v1
+      - uses: guardian/setup-scala@cc7b39e238d789370aa98fa5f07e609b6527657d # v1.1.0
       - uses: scalacenter/sbt-dependency-submission@64084844d2b0a9b6c3765f33acde2fbe3f5ae7d3 # v3.1.0
       
-      - uses: actions/dependency-review-action@v4
+      - uses: actions/dependency-review-action@da24556b548a50705dd671f47852072ea4c105d9 # v4.7.1
         with:
           fail-on-severity: high
           comment-summary-in-pr: always
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,27 @@
+# Avoids introducing vulnerable dependencies to the main branch.
+
+name: Dependency review
+on:
+  pull_request:
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+
+    permissions:
+      # To write dependency submission
+      contents: write
+      # To write comment on PR
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v5
+
+      # Required because Dependabot Alerts doesn't natively support Scala
+      - uses: guardian/setup-scala@v1
+      - uses: scalacenter/sbt-dependency-submission@64084844d2b0a9b6c3765f33acde2fbe3f5ae7d3 # v3.1.0
+      
+      - uses: actions/dependency-review-action@v4
+        with:
+          fail-on-severity: high
+          comment-summary-in-pr: always


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,27 @@
...
+      # To write dependency submission
+      contents: write
+      # To write comment on PR
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v5
+
+      # Required because Dependabot Alerts doesn't natively support Scala
+      - uses: guardian/setup-scala@v1
```
tjsilver, line 21: Could we use the commit SHA for the version here?

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    permissions:
+      # To write dependency submission
+      contents: write
+      # To write comment on PR
+      pull-requests: write
+
+    steps:
+      - uses: actions/checkout@v5
```
tjsilver, line 18: Could we use the commit SHAs for the versions of all the actions in this file, as per the dependency submission workflow?

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      # Required because Dependabot Alerts doesn't natively support Scala
+      - uses: guardian/setup-scala@v1
+      - uses: scalacenter/sbt-dependency-submission@64084844d2b0a9b6c3765f33acde2fbe3f5ae7d3 # v3.1.0
```
tjsilver, line 22: Does the dependency submission workflow need to run here too?

NovemberTang, line 22: Will this overwrite the dependencies on main? Would a custom [correlator ID](https://github.com/scalacenter/sbt-dependency-submission) help?

kelvin-chappell: It shouldn't conflict with the main dependency graph because the default correlator is 
> the concatenation of the workflow name, the job id and the action id
but I'll check this

kelvin-chappell: Unfortunately to compare Scala deps it does because the branch dependency graph to compare with the main dependency graph isn't created automatically

.github/workflows/dependency-review.yml
kelvin-chappell: Done in 5bdf367252f074fbe984cc2d06083b2810857056

tjsilver (approved):
👍

.github/workflows/dependency-review.yml
kelvin-chappell: In order for dependency review to work it has to be able to compare the latest main branch dependency graph (shown in the GitHub console) with the latest PR branch dependency graph, so they can't conflict.  When we get the next Scala change we can see what dependency review says has changed and verify this with the latest main branch dependency graph.

Submission is made [here](https://github.com/scalacenter/sbt-dependency-submission/blob/e4e1a9718ebb2a76347e357b17740e98752b5c11/sbt-plugin/src/main/scala/ch/epfl/scala/SubmitDependencyGraph.scala#L119) and results can be seen [here](https://api.github.com/repos/guardian/janus-app/dependency-graph/snapshots/31868638).

More easy to see changes in [review output in workflow log](https://github.com/guardian/janus-app/actions/runs/16963256915/job/48080861644#step:5:26).

The official documentation around this is very poor.
See:
* [Deduplication of manifests](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/dependency-graph-supported-package-ecosystems#deduplication-of-manifests)
* [API call documentation](https://docs.github.com/en/rest/dependency-graph/dependency-submission?apiVersion=2022-11-28#create-a-snapshot-of-dependencies-for-a-repository) (presumably 'ref' is used to identify a push to the repo default branch but that isn't documented)
* https://github.blog/changelog/2025-05-05-dependency-graph-deduplication-is-now-generally-available/
* 'Submitted dependencies will be shown in dependency review, but are not available in your organization's dependency insights.'
https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/using-the-dependency-submission-api?utm_source=chatgpt.com#about-the-dependency-submission-api
(This actually isn't true if the branch submitted is the default branch.)
* https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#best-practices-for-using-the-dependency-review-api-and-the-dependency-submission-api-together

PR #686, opened 2025-09-29
feat: Add contact link to user account page

kelvin-chappell:
Let people know who to contact if they have problems with passkeys.

Makes it easier to know who to contact when things go wrong.

> [!NOTE]
> Will need to update local dev property file when this has been merged.

### DIFF
--- a/app/conf/Config.scala
+++ b/app/conf/Config.scala
@@ -119,7 +119,9 @@ object Config {
     requiredString(config, "auth.google.2faGroupId")
   }
 
-  /** Link suitable for an HTML anchor href attribute.  E.g. a URL or an email address. */
+  /** Link suitable for an HTML anchor href attribute. E.g. a URL or an email
+    * address.
+    */
   def passkeysManagerLink(config: Configuration): String = {
     requiredString(config, "passkeys.manager.contactLink")
   }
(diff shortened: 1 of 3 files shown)


tjsilver (approved):
👍

app/conf/Config.scala
```diff
@@ -119,6 +119,10 @@ object Config {
     requiredString(config, "auth.google.2faGroupId")
   }
 
+  def devxSecurityContactLink(config: Configuration): String = {
+    requiredString(config, "devxSecurity.contact.link")
```
adamnfish-gu, line 123: Janus-app tries not to contain any Guardian specific information. Can we rename this without our team name?

app/views/userAccount.scala.html
```diff
@@ -52,6 +60,8 @@ <h3 class="header orange-text">Passkeys</h3>
             your AWS accounts.
             One should be on-device (e.g. Mac Touch ID or Windows Hello)
             and the other off-device (e.g. security key or phone).</p>
+        <p>Please contact the <a href="@devxSecurityContactLink">DevX Security team</a> if you have any problems
```
adamnfish-gu, line 63: Here as well, ideally this wouldn't be hard-coded to our organisation. We can talk about this if we want to change approach but I know there are a couple of companies that have used Janus in the past.

kelvin-chappell: As discussed, will replace with separate generic params for the link and for its description.

PR #699, opened 2025-10-24
feat(passkeys): Support password manager browser extensions

kelvin-chappell:
This change enables passkeys to be authenticated by and stored in password managers, as well as the native options provided by the OS and the browser currently.

### Conditions for changed experience
1. A password manager extension is enabled in the browser
2. The user is signed into the password manager
3. They have a passkey for Janus stored in it

### The changed experience
At registration and authentication time, an autofill UI dialog provided by the password manager appears top right instead of the native modal dialog.  (If multiple password managers are installed, a dialog supplied by one of them opens.  I'm not sure what the rules of precedence are here.)

Users can be provided with an org-managed password manager extension to make it easier to provide passkeys.

> [!CAUTION]
> Not all password managers respect the [user verification options of the WebAuthn spec](https://web.dev/articles/webauthn-user-verification) without further configuration.  Even though the authentication response they send to the server for verification claims that user verification took place.
 
## To Test

There are many scenarios but these are the important ones:

| Test | Local Chrome | Local Firefox | Local Safari | Prod Chrome | Prod Firefox | Prod Safari |
|-----|---------|--------|-------|-------|-------|-------|
| Sent straight through to target when opted out of passkeys feature | ✅ | ✅ | ✅ |
| Registered new passkey from password manager | ✅ | ✅ | ✅ |
| Registered new Yubikey passkey | ✅ | ✅ | ✅ |
| Registered new Yubikey passkey when passkey from password manager already registered | ✅ | ✅ | ✅ |
| Authenticated with passkey from password manager | ✅ | ✅ | ✅ |
| Authenticated with Yubikey | ✅ | ✅ | ✅ |
| Authenticated with Yubikey when no password manager enabled | ✅ | ✅ | ✅ |
| Authenticated with Yubikey when password manager enabled but has no passkey for domain | ✅ | ✅ | ✅ |
| Redirected to user account page when no passkeys registered | ✅ | ✅ | ✅ |

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -96,6 +96,7 @@ class AppComponents(context: ApplicationLoader.Context)
     PasskeyAuthenticator.fromResource(
       "passkeys_aaguid_descriptions.json"
     )
+  private val passwordManagers = Set("LastPass")
 
   private val passkeysEnabled: Boolean =
     configuration
@@ -141,7 +142,12 @@ class AppComponents(context: ApplicationLoader.Context)
       host,
       janusData,
       passkeysEnabled,
-      passkeysEnablingCookieName
+      passkeysEnablingCookieName,
+      passkeyAuthenticatorMetadata.collect {
+        case (aaguid, authenticator)
+            if passwordManagers.contains(authenticator.description) =>
+          aaguid
+      }.toSet
     ),
     new Audit(janusData, controllerComponents, authAction),
     new RevokePermissions(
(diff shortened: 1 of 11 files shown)


tjsilver (commented):
I could not test as I was unable to add a passkey to Lastpass for Janus on Firefox. When I tried on prod Janus, adding the passkey to Lastpass extension was successful, but on Janus I got the toast error 'invalid passkey field'. I am also unable to get to the local janus running this branch with the error: `Could not resolve substitution to a value: ${PASSKEY_MANAGER_CONTACT_LINK}`. So I think I'm caught in a situation where I can't register a passkey, but also can't proceed without one on this branch.

PR #808, opened 2026-03-12
Add dev container support

aug24:
Add capability to run in a dev container.


adamnfish-gu (approved):
Minor point about documenting the ports (because it's an unusual layout), but looks great.

PR #823, opened 2026-03-31
Add session tags to role assumption

adamnfish-gu:

Adds tags to the assumed STS session for Janus users, that document the username and permission ID. These tags are marked as "transitive", which means they will also be propogated through subsequent role assumptions.

This adds a few useful benefits:

### Governance

These tags embed useful Janus information into the session, which means the username and permission will be clearly visible in AWS' native security and auditing tooling

### Access control

These tags can be used to enforce tag based access control, which can provide an extra layer of security to operations that should be locked down to a specific permission. If a team wants to be sure that only a specific permission can grant access to a resource in AWS, they'll be able to use the new `gu:janus:permission` tag in an IAM condition. This allows teams to demand explicit access to a resource.

This won't make any difference for normal use but will improve incident response and offers teams additional protections.

### DIFF
--- a/app/aws/Federation.scala
+++ b/app/aws/Federation.scala
@@ -33,6 +33,9 @@ object Federation {
 
   private val awsMinimumSessionLength = 900.seconds
 
+  private val usernameTagKey = "gu:janus:user"
+  private val permissionTagKey = "gu:janus:permission"
+
   private val signInUrl = "https://signin.aws.amazon.com/federation"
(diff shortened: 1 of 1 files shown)


kelvin-chappell (approved):
Nice idea! 👍

adamnfish-gu:
We'll need this PR to get merged before trying this again:

https://github.com/guardian/janus/pull/5428

PR #833, opened 2026-04-13
feat: Extract more detail from managed policies to build developer policies 

kelvin-chappell:
This is to expose the extra data made available in release [v63.1.0](https://github.com/guardian/cdk/releases/tag/v63.1.0) of GuCDK.
The data comes from sections of the managed policy path so the way the path is parsed has changed.
The extra data available is:
1. We can now see which stack a policy is in
2. We can now see which stage a policy gives access to
3. A description is now required, which will help to choose the correct policy for a task 

This extra data will enable us to show more context for the policy in the Janus UI.

## Any additional notes?

There's a fallback to the old way of parsing the policy path so that old policies won't break.
We'll remove this fallback once all the existing policies have been updated to the latest CDK.

Views haven't been touched in this PR.  It just updates the model and view changes will be done separately.

Tested locally with a new manually created policy in the dev playground account alongside old previously working policies.

## See also
* https://github.com/guardian/cdk/pull/2862
* https://trello.com/c/dIKDEITU/4092-make-improvements-to-the-cdk-developer-policy-class


test/logic/AccountsTest.scala
```diff
@@ -92,17 +92,19 @@ class AccountsTest
...
                   .partition("awsPartition")
                   .service("awsService")
                   .resource("awsResource")
                   .build(),
-                "awsResource",
-                s"provisionedRoleId${a.name}",
-                Some(s"description${a.name}"),
-                a
+                policyName = "awsResource",
+                policyGrantId = s"developerPolicyId${a.name}",
```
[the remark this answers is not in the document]
kelvin-chappell: Fixed

app/logic/DeveloperPolicies.scala
```diff
@@ -25,16 +22,28 @@ object DeveloperPolicies {
...
+  ): Option[DeveloperPolicy] = {
+    /* Incoming AWS managed policy path should have structure:
+     * /developer-policy/guardian/<repo-name>/<stack>/<stage>/<grant-id>/
+     */
+    val pathParts = policy.path.split('/')
     for {
-      policyGrantId <- policy.path.split('/').lift(2)
+      // We expect 7 parts including empty string at start
+      pathSections <- Option.when(pathParts.length == 7)(pathParts.tail)
+      stackName <- pathSections.lift(3).filter(!_.isBlank)
```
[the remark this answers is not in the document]
kelvin-chappell: Done

app/models/DeveloperPolicy.scala
```diff
@@ -22,7 +22,9 @@ case class DeveloperPolicy(
     policyArn: Arn,
     policyName: String,
     policyGrantId: String,
-    description: Option[String],
+    stackName: String,
+    stage: String,
+    description: String,
```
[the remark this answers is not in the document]
kelvin-chappell: Done

app/models/DeveloperPolicy.scala
```diff
@@ -31,12 +39,16 @@ object DeveloperPolicy {
       policyArnString: String,
       policyName: String,
       policyGrantId: String,
-      description: Option[String],
+      stackName: String,
```
adamnfish-gu: Why `stackName`? We consistently use stack as the name of the string in our codebases, does it need the `name` suffix?

app/services/DeveloperPolicyCachingService.scala
```diff
@@ -165,10 +165,11 @@ class DeveloperPolicyCachingService(
...
       policy <- IO.fromOption(
-        DeveloperPolicies.toDeveloperPolicy(
-          account,
-          awsPolicy
-        )
+        DeveloperPolicies
+          .toDeveloperPolicy(account, awsPolicy)
+          .orElse(
+            DeveloperPolicies.toDeveloperPolicyFromOldPolicy(account, awsPolicy)
+          )
```
adamnfish-gu: Would it be neater to have this versioning be encapsulated in our `toDeveloperPolicy` logic, where we can test it in one place? Right now it's hard to test this fallback because it's muddled up ion the side effects of this call site. It also means our migration steps require changes in both places.

If we had `toDeveloperPolicy` that tried the new style and falls back to the old, then this part of the code doesn't need to care about how we've achieved that. Given the identical signatures and the fallback content (`unknown`s) in the fallback function, these look like one interface to me!

app/models/DeveloperPolicy.scala
kelvin-chappell: Was just to make it clear it wasn't a multi-property object but will make it consistent

app/services/DeveloperPolicyCachingService.scala
kelvin-chappell: That's true!

app/models/DeveloperPolicy.scala
kelvin-chappell: Renamed in e46ca363620c7b3da7dbc6eb4557838b80161d8a

app/services/DeveloperPolicyCachingService.scala
kelvin-chappell: Reorganised in a7ee36f3714145f697971487274d9d30a106b024

app/models/DeveloperPolicy.scala
```diff
@@ -22,7 +28,9 @@ case class DeveloperPolicy(
     policyArn: Arn,
     policyName: String,
     policyGrantId: String,
-    description: Option[String],
+    stack: String,
+    stage: String,
+    description: String,
```
adamnfish-gu, line 33: Are we sure we want to change the type here?

Having this be "stringly typed" (via the magic string "unknown") loses important information, for example it makes it harder to represent this edge case in the UI.

I think having the others as strings for the duration of the migration away from the old format is fine, in contrast!

I don't feel super strongly, especially if we're pretty sure this can never come up in practice.

app/logic/DeveloperPolicies.scala
```diff
@@ -9,13 +9,28 @@ import models.{
...
 
 import java.net.URLEncoder
+import scala.util.matching.Regex
 
 object DeveloperPolicies {
 
+  /* Matches: /developer-policy/guardian/<repo-name>/<stack>/<stage>/<grant-id>/
+   * Each captured segment must contain at least one non-whitespace character. */
+  private val DeveloperPolicyPath: Regex =
+    """/developer-policy/guardian/[^\s/]+/([^\s/]+)/([^\s/]+)/([^\s/]+)/""".r
```
adamnfish-gu, line 19: What's the reason for not capturing the repository name? No harm, but if it is here it might be useful?

I also just want to check we mean to explicitly disallow `/` characters in the grant-id? At the moment if there are any other `/` characters after this prefix it'll fail to match, and unexpectedly won't show up in Janus.

test/logic/DeveloperPoliciesTest.scala
```diff
@@ -35,92 +31,188 @@ class DeveloperPoliciesTest
...
+        .arn("arn:aws:iam::123:policy/developer-policy/pol")
+        .path("/developer-policy/")
+        .policyName("pol")
+        .build()
+
+      toDeveloperPolicyFromOldPolicy(account, shortPathPolicy) shouldBe None
+    }
+  }
+
+  "toDeveloperPolicyWithFallback" - {
```
adamnfish-gu, line 175: One edge case that emerges from testing the two private pieces separately is what happens between the two of them.

What if there are between 2 and 6 path components? Is the resulting behaviour desired? This is connected to my "what if there are more than 6" comment about whether `/` is allowed in the grant-id part, as well!

app/models/DeveloperPolicy.scala
kelvin-chappell: I'm only expecting this to be the case for a few hours to keep everything working for the time it takes to migrate existing policies so I think it's worth having the expectation that it will be present.

app/logic/DeveloperPolicies.scala
kelvin-chappell: I think if we allow `/` it will become very difficult to troubleshoot missing segments of the path.  Although for consistency we should disallow it in the DeveloperPolicyGrant.id as well along with the characters that AWS won't let us use

test/logic/DeveloperPoliciesTest.scala
kelvin-chappell: I think it makes problems easier to troubleshoot if we're quite prescriptive about the shape of the path.  Telling people the path always has 6 segments and this is what they are is perhaps easier than saying everything from segment 5 on is the grant ID.  This is also the case with the ARN which would then have the policy name on the end.

adamnfish-gu (approved):

kelvin-chappell:
Successfully deployed to Prod and there I can see old and the new test policies:
image

I'm able to get credentials for both old and new policies.
The audit trail shows both too.
So success all round.

PR #847, opened 2026-04-22
refactor: Remove fallback parsing method for old Developer Policies

kelvin-chappell:
This removes some redundant code.  Now that all dev policies are defined using GuCDK [v63.1.0](https://github.com/guardian/cdk/releases/tag/v63.1.0) we don't need a fallback parsing method as none are in the legacy format.

Less code to maintain.  Code fails faster.

## TODO before merging
- [x] Update last legacy dev policy: https://github.com/guardian/devx-backup-monitoring/pull/216


tjsilver (approved):
Looks good!

PR #849, opened 2026-04-27
UI tweaks

adamnfish-gu:
Janus' developer policy feature remains a work in progress, and the UI will evbolve as we see how people use it. This PR makes some changes to the dev policy UI elements ahead of us rolling it out more widely.

Changes how we display developer policies and their grants on permissions pages. We also bring this new visual language to the developer policy status page and the account-specific version of the same information.

The previous design has a few shortcomings:

- The display is a bit broken after the reordering of the dev policies / standard permissions, because the dev policy header runs into the account header. These elements were previously assumed to be non-adjacent.
- Also since the reordering, we lack clear messaging between the developer policy grant sections and the standard permissions. It isn't necessarily clear from the dividing line that we've switched away from the final developer policy, where one or more exist.
- The hover UI for exposing additional dev policy information looks nice, but has a few issues. It requires mouse precision and offers no keyboard fallback, so it isn't super accessible. It also runs off the page at the moment, if the hover button is near the screen's edge.
- There's nowhere on the permissions page where we show the actual developer policy grant ID that bring in the policies that appear.

With these things in mind, this PR tweaks the layout.

### After

Here is a permissions page (admin / home)

Screenshot 2026-04-27 at 15 58 10

Screenshot 2026-04-27 at 16 02 46

... and the same visual language on the policy status pages (handyman icon and the "chips" for policy metadata

Screenshot 2026-04-27 at 16 05 06

### Before

For reference, here's a permissions screen today:

Screenshot 2026-04-27 at 16 06 34

## Any additional notes?

This is tested with the themed diaplay modes. We use the `backdrop-filter: brightness(xx%);` approach to ensure that UI elements look consistent over background elements that will change colour at different times of the year.

The "expand" button is keyboard accessible, and while testing this we also spotted that the favourite buttons had no visual focus indication.


app/logic/ViewHelpers.scala
```diff
@@ -51,4 +51,9 @@ object ViewHelpers {
       .lift(column)
       .getOrElse(Nil)
   }
+
+  enum UIImage {
```
adamnfish-gu, line 55: This is because the material icon set doesn't include a GitHub logo. Our "dev policy metadata chip" supports an image (for GitHub) or an icon (for the other chips).

app/views/fragments/developerPolicyGrants.scala.html
```diff
@@ -1,53 +1,49 @@
...
+                                        </button>
+                                    </div>
+                                </span>
+                            </div>
+                            <div class="card-content--permission-heading__text">
                                 @policyPerm.description
                                 @if(policyPerm.shortTerm) {
                                     <i class="material-icons tooltipped" data-position="bottom" data-delay="50" data-tooltip="This permission only grants short-term access">error</i>
                                 }
-                                <span class="developer-policy-info">
```
adamnfish-gu, line 26: We've swapped this popup for the `developerPolicyInfoChip`s, above.

kelvin-chappell (approved):
A huge improvement! 👍

PR #852, opened 2026-04-29
feat: Add permission type to audit trail

kelvin-chappell:
This adds two changes to the audit trail:
1. There's now a new column that shows `Permission Type`. Possible values are `Account Permission` or `Developer Policy`.  We store this value in the audit DB table.
2. For `Developer Policy` permissions, the `Access` value now links to the AWS console page for the underlying managed policy of the developer policy.

We can see clearly which type of permission has been granted and found out more info about it.

## Images
Screenshot 2026-04-29 at 15 33 48

## Any additional notes?
All previously created rows are labelled as `Account Permission`.  This means a small number of rows will be inaccurate where we started using Developer policies before this PR was deployed.

The audit trail viewer will need to have access to the AWS console for a given account to be able to follow the link.

### DIFF
--- a/app/logic/AuditTrail.scala
+++ b/app/logic/AuditTrail.scala
@@ -1,7 +1,14 @@
 package logic
 
 import com.gu.googleauth.UserIdentity
-import com.gu.janus.model.{ACL, AuditLog, JanusAccessType, Permission}
+import com.gu.janus.model.{
+  ACL,
+  AuditLog,
+  JanusAccessType,
+  Permission,
+  PermissionType
+}
+import com.gu.janus.model.PermissionType.AccountPermission
 import logic.UserAccess.username
 import play.api.Logging
 import software.amazon.awssdk.services.dynamodb.model.AttributeValue
@@ -30,6 +37,7 @@ object AuditTrail extends Logging {
   // Whether access request was for AWS console or credentials for local use
   val accessTypeAttrName = "j_accessType"
   val isExternalAttrName = "j_external"
+  val permissionTypeAttrName = "j_permissionType"
 
   /** Database item attributes for a single audit log entry. */
   case class AuditLogDbEntryAttrs(
@@ -39,7 +47,8 @@ object AuditTrail extends Logging {
       sessionDuration: (String, AttributeValue),
       accessLevel: (String, AttributeValue),
       accessType: (String, AttributeValue),
-      isExternal: (String, AttributeValue)
+      isExternal: (String, AttributeValue),
+      permissionType: (String, AttributeValue)
   ) {
     val toMap: Map[String, AttributeValue] = Seq(
       partitionKey,
@@ -48,7 +57,8 @@ object AuditTrail extends Logging {
       sessionDuration,
       accessLevel,
       accessType,
-      isExternal
+      isExternal,
+      permissionType
     ).toMap
   }
   object AuditLogDbEntryAttrs {
@@ -73,6 +83,9 @@ object AuditTrail extends Logging {
         ),
         isExternal = isExternalAttrName -> AttributeValue.fromN(
           (if (auditLog.external) 1 else 0).toString
+        ),
+        permissionType = permissionTypeAttrName -> AttributeValue.fromS(
+          auditLog.permissionType.serialised
         )
       )
   }
@@ -84,7 +97,8 @@ object AuditTrail extends Logging {
       janusAccessType: JanusAccessType,
       duration: Duration,
       acl: ACL,
-      hasExplicitAccess: Boolean
+      hasExplicitAccess: Boolean,
+      permissionType: PermissionType
   ): AuditLog =
     AuditLog(
       permission.account.authConfigKey,
@@ -93,7 +107,8 @@ object AuditTrail extends Logging {
       duration,
       permission.label,
       janusAccessType,
-      !hasExplicitAccess
+      !hasExplicitAccess,
+      permissionType
     )
 
   /** Extract nice error message from db conversion.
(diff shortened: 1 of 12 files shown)


app/logic/AuditTrail.scala
```diff
@@ -149,15 +164,21 @@ object AuditTrail extends Logging {
...
-      dateTime,
-      duration,
-      accessLevel,
-      accessType,
-      external
-    )
+    } yield {
+      val permissionType = stringValue(attrs, permissionTypeAttrName)
+        .flatMap(PermissionType.fromString)
+        .getOrElse(AccountPermission)
```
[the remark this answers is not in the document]
kelvin-chappell: Don't think this is necessary

app/views/audit.scala.html
```diff
@@ -56,7 +57,17 @@ <h1 class="header orange-text">Audit trail</h1>
                                     <a href="/audit/user/@auditLog.username">@auditLog.username</a>
                                 }
                             </td>
-                            <td>@auditLog.accessLevel</td>
+                            <td>
+                                @if(auditLog.permissionType == PermissionType.DeveloperPolicyPermission){
+                                    @developerPolicyLinks.get(auditLog.accessLevel).map { url =>
+                                        <a href="@url">@auditLog.accessLevel.stripPrefix("iam-")</a>
```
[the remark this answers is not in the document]
kelvin-chappell: I don't think different accounts can have the same developer policy slug

app/views/audit.scala.html
```diff
@@ -56,7 +57,17 @@ <h1 class="header orange-text">Audit trail</h1>
...
+                                    }.getOrElse {
+                                        @auditLog.accessLevel.stripPrefix("iam-")
+                                    }
```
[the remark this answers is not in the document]
kelvin-chappell: Showing a URL-decoded value in 0125ae86f51e837ca71fd42447caab7bfdec6c69

app/logic/DeveloperPolicies.scala
```diff
@@ -82,6 +82,16 @@ object DeveloperPolicies {
...
 
+  /** Builds a map from each policy's access-level slug to its AWS console URL.
+    * Used by the audit view to render links for developer policy rows.
+    */
+  def developerPolicyLinksBySlug(
+      policies: Set[DeveloperPolicy]
+  ): Map[String, String] =
+    policies
+      .map(p => developerPolicySlug(p.policyName) -> developerPolicyLink(p))
+      .toMap
```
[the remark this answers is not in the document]
kelvin-chappell: I think this is very unlikely to happen and we'll address it when it does.

test/logic/AuditTrailTest.scala
```diff
@@ -102,16 +108,47 @@ class AuditTrailTest
...
+        accountPartitionKeyName -> AttributeValue.fromS("account"),
+        userNameAttrName -> AttributeValue.fromS("username"),
+        timestampSortKeyName -> AttributeValue.fromN("1446650520000"),
+        durationAttrName -> AttributeValue.fromN("3600"),
+        accessLevelAttrName -> AttributeValue.fromS("dev"),
+        accessTypeAttrName -> AttributeValue.fromS("console"),
+        isExternalAttrName -> AttributeValue.fromN("1")
+      )
+
+      "defaults to AccountPermission for backwards compatibility" in {
```
adamnfish-gu, line 146: Great test, nice one.

app/views/audit.scala.html
```diff
@@ -56,7 +58,17 @@ <h1 class="header orange-text">Audit trail</h1>
                                     <a href="/audit/user/@auditLog.username">@auditLog.username</a>
                                 }
                             </td>
-                            <td>@auditLog.accessLevel</td>
+                            <td>
+                                @if(auditLog.permissionType == PermissionType.DeveloperPolicyPermission){
+                                    @developerPolicyLinks.get(auditLog.accessLevel).map { url =>
```
adamnfish-gu: This get (and the `getOrElse` below) highlight that it's a shame to pass this link information in a separate parameter and have to re-assemble it. I see that the reason is we've lost the ARN when we get to the audit log, and need to look it up from the dev policy cache.

Another approach might be to do the recombination in the controller and pass the "hydrated" data to the template. I'm not sure if it's any easier though, I guess we're just in a bad spot either way, here.

However, as mentioned, I'm not certain we need to display these policies as links in the audit. That might let us jettison some of this complexity?

app/views/audit.scala.html
```diff
@@ -56,7 +58,17 @@ <h1 class="header orange-text">Audit trail</h1>
                                     <a href="/audit/user/@auditLog.username">@auditLog.username</a>
                                 }
                             </td>
-                            <td>@auditLog.accessLevel</td>
+                            <td>
+                                @if(auditLog.permissionType == PermissionType.DeveloperPolicyPermission){
```
adamnfish-gu, line 62: Would a match make it clearer the two cases we're handling here? We clearly show the `DeveloperPolicyPermission` but the `AxxountPermission` is just an `else` block. I don't mind and the extra `{}` twirl requires for matches is mildly annoying.

This is a super minor point!

app/views/audit.scala.html
```diff
@@ -56,7 +58,17 @@ <h1 class="header orange-text">Audit trail</h1>
...
+                                    @developerPolicyLinks.get(auditLog.accessLevel).map { url =>
+                                        <a href="@url">@developerPolicyDisplayName(auditLog.accessLevel)</a>
```
adamnfish-gu: I think the table is going to very squashed with these new fields. I wonder if there's a different way to convey the policy name, and if we could use an icon for the type column (or omit it)?

No dramas though, let's get it out and have a look at the result.

adamnfish-gu (approved):
👍  Look ok, but there are lots of optional thoughts! Approved with the caveat we might be able to make this a bucnh simpler.

-------

Some of the complexity here comes from us linking to the IAM policy, because we need to look this information up from the controller and then tie it all back together when we render it.

There's also a UX problem whenever we link to a resource in AWS from Janus (the policies status page has this problem as well).

But if we consider the use case here, the audit trail caters to an audience that aren't necessarily expected to have access to a dev policy in an AWS account.

Give all these things together, I wonder if we could drop the link requirement to:
- remove the re-combining logic with its link generation
- change the display of the dev policy name so it's a bit less prominent

For the latter, we're using a normal text "prose" font size and style for a policy name. This makes sense when it's a link, but we're displaying a policy name (and might use a monospace font for that) and taking up a lot of horizontal space as a result (we'd probably use a smaller font for the policy name if we were designing it from scratch, like we do elsewhere in Janus).

See what you think!

app/views/audit.scala.html
kelvin-chappell: We could potentially just show an icon next to the access entry but I'll get this out as is (minus the links to AWS) and we can think about that separately.

PR #854, opened 2026-05-04
Add metrics

aug24:

Emit metrics on four events:
 * successful creation of credentials emits packed size with dimensions
 * failed requests (credentials, multi-credentials, console)

This change will give us information about how often 
 * requests are denied
 * requests fail
 * developer policies and "traditional" policies are used
 * the packed size of the requests.

This last is important as some of our larger policies are close to the limit.

## Any additional notes?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1214262369447238

### DIFF
--- a/app/aws/Clients.scala
+++ b/app/aws/Clients.scala
@@ -5,6 +5,7 @@ import conf.Config
 import play.api.Configuration
 import software.amazon.awssdk.auth.credentials.*
 import software.amazon.awssdk.regions.Region.EU_WEST_1
+import software.amazon.awssdk.services.cloudwatch.CloudWatchClient
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
 import software.amazon.awssdk.services.iam.IamClient
 import software.amazon.awssdk.services.ssm.SsmClient
@@ -57,6 +58,13 @@ object Clients {
     .credentialsProvider(securityCredentialsProviderChain)
     .region(EU_WEST_1)
     .build()
+
+  lazy val cloudwatchClient: CloudWatchClient =
+    CloudWatchClient
+      .builder()
+      .credentialsProvider(securityCredentialsProviderChain)
+      .region(EU_WEST_1)
+      .build()
 
   /** Gives an IAM client scoped to the given account so that it can look for
     * [[com.gu.janus.model.ProvisionedRole]] data.
--- a/app/aws/Federation.scala
+++ b/app/aws/Federation.scala
@@ -94,7 +94,7 @@ object Federation {
       permission: Permission,
       sts: StsClient,
       duration: Duration
-  ): Credentials = {
+  ): (Credentials, Integer) = {
     val requestBuilder = AssumeRoleRequest
       .builder()
       .roleArn(roleArn)
@@ -112,7 +112,7 @@ object Federation {
     }
     val request = requestBuilder.build()
     val response = sts.assumeRole(request)
-    response.credentials()
+    (response.credentials(), response.packedPolicySize())
   }
 
   def generateLoginUrl(
@@ -159,13 +159,14 @@ object Federation {
     }
 
     // assume role in the target account to authenticate the revocation
-    val creds = Federation.assumeRole(
-      username,
-      roleArn,
-      Policies.revokeAccessPermission(account),
-      stsClient,
-      Federation.awsMinimumSessionLength
-    )
+    val (creds, _) = Federation
+      .assumeRole(
+        username,
+        roleArn,
+        Policies.revokeAccessPermission(account),
+        stsClient,
+        Federation.awsMinimumSessionLength
+      )
     val sessionCredentials = AwsSessionCredentials.create(
       creds.accessKeyId,
       creds.secretAccessKey,
--- a/app/controllers/Janus.scala
+++ b/app/controllers/Janus.scala
@@ -374,13 +374,21 @@ class Janus(
           )
           Some((credentials, permission))
         } catch {
-          case NonFatal(e) =>
+          case e: PackedPolicyTooLargeException =>
             logger.error("Exception creating credentials", e)
-            metricsService.putExceptionOnRequest(
+            metricsService.putTooLargeRequest(
               permissionId,
               accessType,
               accessSource,
               e
+            )
+            throw e
+          case NonFatal(e) =>
+            logger.error("Exception creating credentials", e)
+            metricsService.putFailedRequest(
+              permissionId,
+              accessType,
+              accessSource
             )
             throw e
         }
--- a/app/controllers/Janus.scala
+++ b/app/controllers/Janus.scala
@@ -14,13 +14,21 @@ import models.AccessSource.Internal
 import models.{AccountAccess, DeveloperPolicy, PasskeyAuthenticator}
 import play.api.mvc.*
 import play.api.{Configuration, Logging, Mode}
-import services.{DeveloperPolicyFinder, DeveloperPolicyStatusManager}
+import services.{
+  DeveloperPolicyFinder,
+  DeveloperPolicyStatusManager,
+  MetricsService
+}
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
 import software.amazon.awssdk.services.sts.StsClient
-import software.amazon.awssdk.services.sts.model.Credentials
+import software.amazon.awssdk.services.sts.model.{
+  Credentials,
+  PackedPolicyTooLargeException
+}
 
 import java.time.*
 import java.time.format.DateTimeFormatter
+import scala.util.control.NonFatal
 
 class Janus(
     janusData: JanusData,
@@ -33,8 +41,9 @@ class Janus(
     passkeysEnablingCookieName: String,
     passkeyAuthenticatorMetadata: Map[AAGUID, PasskeyAuthenticator],
     developerPolicyService: DeveloperPolicyFinder
-      with DeveloperPolicyStatusManager
-)(using dynamodDB: DynamoDbClient, mode: Mode, assetsFinder: AssetsFinder)
+      with DeveloperPolicyStatusManager,
+    metricsService: MetricsService
+)(using dynamoDB: DynamoDbClient, mode: Mode, assetsFinder: AssetsFinder)
     extends AbstractController(controllerComponents)
     with ResultHandler
     with Logging {
@@ -232,7 +241,7 @@ class Janus(
           .withHeaders(CACHE_CONTROL -> "no-cache")
       }) getOrElse {
         logger.warn(
-          s"console login to $permissionId denied for ${username(request.user)}"
+          s"console url login to $permissionId denied for ${username(request.user)}"
         )
         Forbidden(views.html.permissionDenied(request.user, janusData))
       }
@@ -287,7 +296,7 @@ class Janus(
           .withHeaders(CACHE_CONTROL -> "no-cache")
       }) getOrElse {
         logger.warn(
-          s"denied credentials to $rawPermissionIds for ${username(request.user)}"
+          s"denied multi credentials to $rawPermissionIds for ${username(request.user)}"
         )
         Forbidden(views.html.permissionDenied(request.user, janusData))
       }
@@ -317,44 +326,70 @@ class Janus(
       durationParams: (Option[Duration], Option[ZoneId]),
       developerPolicies: Set[DeveloperPolicy]
   ): Option[(Credentials, Permission)] = {
-    val (requestedDuration, tzOffset) = durationParams
-    for {
-      (permission, accessSource, permissionType) <-
-        checkUserPermissionWithSource(
-          username(user),
+    checkUserPermissionWithSource(
+      username(user),
+      permissionId,
+      Instant.now(),
+      janusData,
+      developerPolicies
+    ) match {
+      case Some(permission, accessSource, permissionType) =>
+        try {
+          val (requestedDuration, tzOffset) = durationParams
+
+          val duration = Federation.duration(
+            permission,
+            requestedDuration,
+            tzOffset.map(Clock.system)
+          )
+
+          val roleArn =
+            Config.roleArn(permission.account.authConfigKey, configuration)
+
+          val (credentials, size) = Federation.assumeRole(
+            username(user),
+            roleArn,
+            permission,
+            stsClient,
+            duration
+          )
+          val auditLog = AuditTrail.createLog(
+            user,
+            permission,
+            accessType,
+            duration,
+            janusData.access,
+            accessSource == Internal,
+            permissionType
+          )
+          AuditTrailDB.insert(auditLog)
+          logger.info(
+            s"$accessType access to $permissionId granted for ${username(user)}"
+          )
+          metricsService.putSuccessfulRequest(
+            permissionId,
+            accessType,
+            accessSource,
+            size
+          )
+          Some((credentials, permission))
+        } catch {
+          case NonFatal(e) =>
+            logger.error("Exception creating credentials", e)
+            metricsService.putExceptionOnRequest(
+              permissionId,
+              accessType,
+              accessSource,
+              e
+            )
+            throw e
+        }
+      case None =>
+        metricsService.putDeniedRequest(
           permissionId,
-          Instant.now(),
-          janusData,
-          developerPolicies
+          accessType
         )
-      duration = Federation.duration(
-        permission,
-        requestedDuration,
-        tzOffset.map(Clock.system)
-      )
-      roleArn = Config.roleArn(permission.account.authConfigKey, configuration)
-      credentials = Federation.assumeRole(
-        username(user),
-        roleArn,
-        permission,
-        stsClient,
-        duration
-      )
-      auditLog = AuditTrail.createLog(
-        user,
-        permission,
-        accessType,
-        duration,
-        janusData.access,
-        accessSource == Internal,
-        permissionType
-      )
-      _ = AuditTrailDB.insert(auditLog)
-    } yield {
-      logger.info(
-        s"$accessType access to $permissionId granted for ${username(user)}"
-      )
-      (credentials, permission)
+        None
     }
   }
 
--- a/app/services/MetricsService.scala
+++ b/app/services/MetricsService.scala
@@ -1,0 +1,248 @@
+package services
+
+import aws.Clients.cloudwatchClient
+import com.gu.janus.model.JanusAccessType
+import models.AccessSource
+import play.api.Logging
+import services.MetricsService.*
+import software.amazon.awssdk.services.cloudwatch.model.*
+import software.amazon.awssdk.services.sts.model.PackedPolicyTooLargeException
+
+import scala.jdk.CollectionConverters.*
+
+class MetricsService(
+    val stage: String,
+    val stack: String = "security",
+    val app: String = "janus"
+) extends Logging {
+
+  private val namespace = s"/$stage/$stack/$app"
+
+  def putDeniedRequest(
+      permissionId: String,
+      accessType: JanusAccessType
+  ): Unit = {
+    try {
+      val deniedMetricRequest =
+        getDeniedMetricRequest(permissionId, accessType)
+      cloudwatchClient.putMetricData(deniedMetricRequest)
+    } catch {
+      case e: Throwable =>
+        logger.error(s"Unable to send denied request metric", e);
+    }
+  }
+
+  private[services] def getDeniedMetricRequest(
+      permissionId: String,
+      accessType: JanusAccessType
+  ): PutMetricDataRequest = PutMetricDataRequest
+    .builder()
+    .namespace(namespace)
+    .metricData(
+      MetricDatum
+        .builder()
+        .metricName(deniedRequestMetricName)
+        .value(1)
+        .unit(StandardUnit.COUNT)
+        .dimensions(
+          Set(
+            getPermissionIdDimension(permissionId),
+            getAccessTypeDimension(accessType)
+          ).toList.asJava
+        )
+        .build()
+    )
+    .build()
+
+  private def putFailedRequest(
+      permissionId: String,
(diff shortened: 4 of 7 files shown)


app/aws/Federation.scala
```diff
@@ -94,7 +94,7 @@ object Federation {
       permission: Permission,
       sts: StsClient,
       duration: Duration
-  ): Credentials = {
+  ): AssumeRoleResponse = {
```
aug24: Because the packed size is a val on the response, rather than add cloudwatch to this class, it was cleaner to return the response and keep all the cloudwatch work in Janus.scala

app/aws/Clients.scala
```diff
@@ -58,6 +59,13 @@ object Clients {
     .region(EU_WEST_1)
     .build()
 
+  lazy val cloudwatchAsyncClient: CloudWatchAsyncClient =
```
aug24: Add the metrics client

app/controllers/Janus.scala
```diff
@@ -317,45 +324,101 @@ class Janus(
...
+    val policyMetricDimensions: Map[String, String] = convertToDimensions(
+      developerPolicies
+    ) + ("permissionId" -> permissionId)
+      + ("accessType" -> accessType.toString)
+
+    try {
+      checkUserPermissionWithSource(
         username(user),
         permissionId,
-        Instant.now(),
```
aug24: This was passed without alteration down a couple of layers to UserAccess and then used in exactly one location.  I moved it to that loctaion.

app/aws/Clients.scala
adamnfish-gu: Let's not bring Futures into the codebase. The rest of these clients are syncronous and we also already have `IO` in the codebase for some of the Dev Policy work. Properly wiring Future up will mean one of the following:
- a lot of other methods also suddenly end up being Futures as well, not desirable
- we Await the Future result, in which case we should have just used the sync client in the first place

The sync client will be fine here, taken with the caveat about making sure this is resistant to failure. That means ensuring the call is robust to failures and has sensible timeouts set.

app/aws/CloudWatch.scala
```diff
@@ -0,0 +1,72 @@
...
+      value: Int = 1
+  ): Unit = {
+    val dimensions = dimensionStrings.map { case (n, v) =>
+      Dimension.builder().name(n).value(v).build()
+    }.toList
+    val putMetricDataRequest =
+      CloudWatch.buildPutMetricRequest(namespace, metric, value, dimensions)
+    cloudwatchAsyncClient
+      .putMetricData(putMetricDataRequest)
+      .asScala // Future, but we fire and forget
```
adamnfish-gu: As above, I think we can avoid using Future here at all.

If we were, we would certainly not want to fire and forget like this! What about error handling, logging, and so on. We need to consider the "unhappy path" as well. This will be even more important when this is properly wired in, because we will want to make explicitly certain that we continue to serve AWS sessions even if these metrics calls fail.

app/aws/CloudWatch.scala
```diff
@@ -0,0 +1,72 @@
...
+      SuccessfulRequestPolicySizeMetric
+    )
+}
+
+class CloudWatch(val stage: String, val stack: String, val app: String)
+    extends Logging {
+
+  private val namespace = s"/$stage/$stack/$app"
+
+  def put(
```
adamnfish-gu: I'm not sure this function really does anything? We pass in a largely untyped map, so we're essentially duplicating the interface that AWS exposes, and this function's job is just to handle that extra conversion step.

What if we instead expose a type-safe method for each of the three metrics we actually put. Right now each call site is being asked to select the correct use-case specific `CloudWatchMetrics` instance, and build a "stringly typed" Map of the correct data.

With use-case specific methods, the enum above can go away completely, the three call sites become simpler (and more type-safe), and this AWS wrapper is replaced by three simple methods that have the simple and direct task of submitting their CloudWatch metric. An optional private shared method might exist, but only if it is needed - putting a CloudWatch metric isn't a complex task so there's likely no need for the indirection of abstraction here.

app/services/MetricsService.scala
```diff
@@ -0,0 +1,72 @@
...
+    val putMetricDataRequest =
+      CloudWatch.buildPutMetricRequest(namespace, metric, value, dimensions)
+    cloudwatchAsyncClient
+      .putMetricData(putMetricDataRequest)
+      .asScala // Future, but we fire and forget
+  }
+}
+
+object CloudWatch {
+  def buildPutMetricRequest(
```
adamnfish-gu: Here again, this looks very tightly coupled with the above, and the complexity it is solving was all invented by us by having to go through our own AWS wrappers. If we do three explicit metric puts (one for each case) I think the need for this sort of thing probably goes away.

app/aws/Federation.scala
adamnfish-gu: I don't love the idea that we leak the AWS response object into our codebase. It may contain nulls, it is mutable, and it has a bunch of information we don't need on it.

This is used in two places and only one of those uses the extra packed size field. I think returning a Tuple might be fine, but a case class wrapper is likely fine too if you prefer.

app/controllers/Janus.scala
```diff
@@ -42,6 +50,8 @@ class Janus(
   import logic.AccountOrdering.*
   import logic.UserAccess.*
 
+  val cloudWatch: CloudWatch = new CloudWatch(stage, "janus", "janus-app")
```
adamnfish-gu: These values are wrong. The stack is `security` and the app is `janus`.

If we were creating a generic CloudWatch class that encapsulates some CloudWatch information (e.g. the app's identity for the namespace), I'd do this at the "edge" in AppComponents. Hard-coding these values in a controller feels like the wrong place. This would mean our controller takes the metrics class as an argument, instead of taking the stage like it does now.

This might also encourage us to move some of the metrics logic out of being inline in the controller. A shared metrics service that exposes tailored (and bullet-proof) metrics functionality would limit the complexity bump we see in this controller.

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
...
+            user,
+            permissionId,
+            accessType,
+            durationParams,
+            permission,
+            accessSource
+          )
+          cloudWatch.put(
+            SuccessfulRequestPolicySizeMetric,
+            policyMetricDimensions + ("label" -> permission.label) + ("accessSource" -> accessSource.toString),
```
adamnfish-gu: I'm not sure we need the label? This is a human readable version of the ID that's used in the UI, I'm not sure we need to be shipping that string into our metrics?

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
...
+        case Some(permission, accessSource) =>
+          val assumeRoleResponse = invokeAssumeRole(
+            user,
+            permissionId,
+            accessType,
+            durationParams,
+            permission,
+            accessSource
+          )
+          cloudWatch.put(
```
adamnfish-gu: A tailored and typesafe version of this method would make this clearer. Something like this:

```scala
metrics.putSuccessfulRoleAssumption(
  packedPolicySize,
  permissionId,
  accessSource,
  accessType
)
```

We don't need to pass the stage, as mentioned elsewhere. We're also keeping our domain types (AccessSource, and JanusAccessType) while in our domain, which is what we want to be doing.

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
...
+            SuccessfulRequestPolicySizeMetric,
+            policyMetricDimensions + ("label" -> permission.label) + ("accessSource" -> accessSource.toString),
+            assumeRoleResponse.packedPolicySize()
+          )
+          Some(assumeRoleResponse.credentials(), permission)
+        case None =>
+          cloudWatch.put(
```
adamnfish-gu: here as well, it's simple and more type-safe to write this directly. Of course, we don't have the `packedPolicySize` or `accessSource` available if there was no access, but we can include the `accessType` here for consistency with the success metric, above.

```scala
metrics.putDeniedRoleAssumption(
  permissionId,
  accessType
)
```

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
       durationParams: (Option[Duration], Option[ZoneId]),
       developerPolicies: Set[DeveloperPolicy]
   ): Option[(Credentials, Permission)] = {
-    val (requestedDuration, tzOffset) = durationParams
-    for {
-      (permission, accessSource) <- checkUserPermissionWithSource(
+
+    // Metric dimensions
+    val policyMetricDimensions: Map[String, String] = convertToDimensions(
+      developerPolicies
```
adamnfish-gu, line 334: I'm not sure why we're creating dimensions for developer policies here?

This argument is a full copy of the dev policy cache, which is required to be able to check access requirements. Adding every developer policy to each metrics call is almost certainly not what is intended, here?

app/controllers/Janus.scala
adamnfish-gu: The serialisation of the full developer polices cache above is probably not intended, but perhaps the goal was to include the Grant ID that allowed the end-user to obtain this access? If that's the case we'd need to refactor the `checkUserPermissionWithSource` method to return this information. If it's required, let's do that as a follow-up so we are separating that logic change from the work here to add metrics support.

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
...
+        case None =>
+          cloudWatch.put(
+            DeniedRequest,
+            policyMetricDimensions
+          )
+          None
+      }
+    } catch {
+      case NonFatal(e) =>
+        cloudWatch.put(
```
adamnfish-gu: Here as well, it's simple and more type-safe to write this directly. Of course, we don't have the some fields available if the call failed, but we can include the accessType here for consistency with the success metric, above.

```scala
metrics.putDeniedRoleAssumption(
  permissionId,
  accessType
)
```

Might this include the exception type as well?

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
...
+          FailedRequest,
+          policyMetricDimensions
+        )
+        throw e
     }
+
   }
 
+  private def invokeAssumeRole(
```
adamnfish-gu: I think the existence of this function is a sign of trouble.

This function is called once, right next door. That means this function is probably an admission that the calling function has got too big, but by moving a single large and tangled block of inline code elsewhere we haven't actually solved the problem that triggered that observation. Specifically, we've added the indirection of jumping to another function, without doing any simplification.

Let's inline this again and then when everything's in place we can take another look at it. With some of the other changes mentioned here we may decide it is ok for now. If we don't, a more robust fix would be to treat this as a sub-program. We'd write it the same way we do our controllers as a list of separate effectful steps, with each dedicated to a single purpose. I think a refactor of that key "hot path" would be wisely done as a separate  and focused PR though, let's aim to review the inline version first.

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
...
+      accessSource == Internal
+    )
+    AuditTrailDB.insert(auditLog)
+    logger.info(
+      s"$accessType access to $permissionId granted for ${username(user)}"
+    )
+    response
+  }
+
+  private def convertToDimensions(
```
adamnfish-gu: I don't think this is required, as above. Even if it were, I don't hink this would belong in the controller - our tailored cloudwatch put methods would take our domain types (here `Set[DeveloperPolicy]`) and pass them to AWS as needed.

Again, this is only mentioned for reference because this is serialising the entire dev policy cache and should be removed!

test/aws/CloudWatchTest.scala
```diff
@@ -0,0 +1,55 @@
...
+import org.scalatest.matchers.should.Matchers
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks.*
+import software.amazon.awssdk.services.cloudwatch.model.{
+  Dimension,
+  PutMetricDataRequest
+}
+import scala.jdk.CollectionConverters._
+
+class CloudWatchTest
```
adamnfish-gu: I haven't reviewed this because I think it is going to change quite a bit.

app/controllers/Janus.scala
```diff
@@ -317,45 +327,102 @@ class Janus(
...
-    val (requestedDuration, tzOffset) = durationParams
-    for {
-      (permission, accessSource) <- checkUserPermissionWithSource(
+
+    // Metric dimensions
+    val policyMetricDimensions: Map[String, String] = convertToDimensions(
+      developerPolicies
+    ) + ("permissionId" -> permissionId)
+      + ("accessType" -> accessType.toString)
+      + ("stage" -> stage)
```
adamnfish-gu: I don't think the controller needs to pass the stage here. The CloudWatch class already encapsulates the stage as part of building the namespace.

This may change around as well.

app/logic/UserAccess.scala
```diff
@@ -216,9 +215,9 @@ object UserAccess {
 
     val support =
       supportData
-        .map((acl, when) =>
+        .map(acl =>
           val perms =
-            userSupportAccess(username, when, acl).getOrElse(Set.empty)
+            userSupportAccess(username, Instant.now(), acl).getOrElse(Set.empty)
```
adamnfish-gu: Why have we made this change here? I think the existing logic was correct and important.

There are two reasons for the old approach:
- We don't want to do the side effect of reading the current time within our logic - this is a tricky piece of logic and it helps us with our testing to be able to make pure assertions
- This introduces a race-hazard. Another goal of the previous signature was to use a single canonical reference of "now" for all our access checks. This matters for support access, which is time-based.

app/aws/CloudWatch.scala
```diff
@@ -0,0 +1,72 @@
+package aws
+
+import aws.Clients.cloudwatchAsyncClient
+import play.api.Logging
+import software.amazon.awssdk.services.cloudwatch.model.*
+
+import scala.jdk.CollectionConverters.*
+import scala.jdk.FutureConverters.*
+
+sealed class CloudWatchMetric(val name: String, val unit: StandardUnit)
```
adamnfish-gu: This is Scala 3, let's use an enum for ADTs like this. Even in Scala 2 we'd probably use a sealed trait rather than a sealed base class, since we don't need to instantiate `CloudWatchMetric` directly.

Also this is named very generally, but is implemented to specifically manage the metrics for the assume role call. Perhaps this could be RoleAssumptionMetrics?

adamnfish-gu: Having dug in, I now think this type is probably redundant.

The design now is to offer a generic put interface that provides a general CloudWatch interface, and to use this enum to document how it can be used. This doesn't actually bring any type-safety to the dimension or value arguments, so it looks like it's sitting halfway between "just use cloudwatch directly" and "provide a typed interface for our desired functionality".

I think the latter would be a better approach, because it allows us to encapsulate our CloudWatch logic. An alternative design would be to provide a "metrics service" to the Janus controller, that exposes functions it can use for each of its three use-cases.

Once we have that service, let's think about how we want DEV to work. If we want it to be sending real metrics to AWS (I think this is fine), then a `MetricsService` class makes sense. If we don't want that, then a `MetricsService` trait will be a better approach, that uses ` CloudWatchMetrics` production implementation and provides an alternative implementation for dev (e.g. one that logs instead).

Right now CloudWatch functionality is spread over this type, the `CloudWatch` class below, the `CloudWatch` companion object, private functionality embedded in the Janus controller (`convertToDimensions`), and the inline logic in the Janus Controller (some of that is inherent, of course). It also demands we wire the stage into our Janus controller, only to have it be passed to our CloudWatch functions.

The main reasons for this suggestion are to consolidate the CloudWatch logic, and to let the Janus Controller remain focused on the core work that Janus does.

adamnfish-gu (changes requested):
A few specific important points, but I'm also wondering about the structure! Let's chat if I can help with any of this.

app/controllers/Janus.scala
aug24: Hell no.  I had misread that as the dev policies being requested.  Thank you.

app/controllers/Janus.scala
aug24: That's exactly what I thought I was doing, wrongly, and exactly what I think we need.  If teams adjust their policies we will need a heads up (ie alert) for any that are getting too big.

In fact, I think we will need to handle the "too big" case explicitly because we no longer have control over the content so we will have to inform the team why it's not working via the UI.

app/controllers/Janus.scala
aug24: Denied it just where we don't have a match - there's no exception, the user just doesn't have this.  I don't think in usual practice the UI would permit it, but there's always bookmarks.

app/controllers/Janus.scala
aug24: I've re-inlined it, but it does make a very large and hard to read method IMO.

app/logic/UserAccess.scala
aug24: I've reverted it, but I think it may need a review because as far as I can see, the value isn't used in multiple places.

app/aws/CloudWatch.scala
aug24: Agreed.  I was actually trying to follow the pattern from elsewhere in Guardian repos on the expectation that was a preferred style:
https://github.com/guardian/amiable/blob/main/app/metrics/CloudWatch.scala#L23

app/services/MetricsService.scala
```diff
@@ -0,0 +1,248 @@
...
+    )
+    .build()
+
+  private def getAccessSourceDimension(accessSource: AccessSource) = Dimension
+    .builder()
+    .name(accessSourceDimensionName)
+    .value(accessSource.toString)
+    .build()
+
+  def putExceptionOnRequest(
```
adamnfish-gu: Catching here just to re-throw seems very strange!

We're already matching exceptions at the call site, so one argument is that we should be matchign the exception types there. I can see the argument to encapsulate that here as well, but if we're leaving this logic here let's not do the side-effect of re-throwing here. The actual control flow should stay at the call site, in FP terminology we're just "tapping" the exception here to emit a metric.

test/services/MetricsServiceTest.scala
```diff
@@ -0,0 +1,290 @@
...
+          accessType,
+          accessSource,
+          PackedPolicyTooLargeException
+            .builder()
+            .message(s"Packed policy consumes $size% of allotted space")
+            .build()
+        )
+      )
+
+    "should generate exactly correct objects" in {
```
adamnfish-gu: This has been through a bunch of changes already and I don't feel super strongly, but tests like these are often frustrating to maintain. These tests are basically hand-written "snapshot tests". They test the source code as it is currently written, and will fail if that changes. They won't give any helpful feedback when they fail, and these tests don't document any of our code's properties (or the important decisions behind them).

So the two issues I'm talking about here are:

1. The choice of what to test (everything vs focused properties)
2. The fact that these are all bundled into a single test (vs having focused tests with a test message that documents the intent)

Taking the example we discussed - we want to ensure the namespace is consistent, and correct. A focused test with a message explaining that requirement, that fails helpfully when the property is broken, would be much more illuminating for someone that follows. This is a piece of logic to check, because we're threading the three class params through into a shared val that needs to be correctly applied to the request - and it is essential this remains consistent.

In contrast:

```scala
    metric.unit() shouldBe StandardUnit.PERCENT
```

This property says we should use percent as the unit, but this decision is already directly documented in the source code.

I knwo we've talked about this before, but the rest of the codebase tries to group tests together (by function, and then by assumption), and aim for a single assertion per test.

```scala
"function" - {
  "assumption" - {
    val sharedSetup = ???

    "clear description of what we expect" in {
      val result = function(sharedSetup, ...)
      <property-of-result> should <assertion>
    }

    "clear description of another property" in {
      val result = function(sharedSetup, ...)
      <property-of-result> should <assertion>
    }
  }
}
```

This encourages us to choose which properties we want to test, and means when things fail we get detailed breakdown of every failure - not just the first failed assertion in each test.

Have a think about whether you really want to assert the exact value of every property, and even if we do, could we break this up? `"dimensions get properly serialised"`, `"the namespace is should start with the provided stage"`, this sort of thing.

See what you think, no dramas!

app/services/MetricsService.scala
```diff
@@ -0,0 +1,248 @@
...
+import services.MetricsService.*
+import software.amazon.awssdk.services.cloudwatch.model.*
+import software.amazon.awssdk.services.sts.model.PackedPolicyTooLargeException
+
+import scala.jdk.CollectionConverters.*
+
+class MetricsService(
+    val stage: String,
+    val stack: String = "security",
+    val app: String = "janus"
```
kelvin-chappell: Are these ever going to have different values?  Might be better stored in the companion object.

app/aws/Federation.scala
```diff
@@ -94,7 +94,7 @@ object Federation {
       permission: Permission,
       sts: StsClient,
       duration: Duration
-  ): Credentials = {
+  ): (Credentials, Integer) = {
```
kelvin-chappell, line 97: Would it be better to return the whole AssumeRoleResponse here?  Then it would be clearer at the call site what's being returned.  I suppose it's swings and roundabouts.  If that seems like exposing too much information, perhaps returning a small case class instead would be good.

aug24: I originally returned the response, but this was preferred.  I don't think it's worth a case class for a single usage interface.

app/services/MetricsService.scala
aug24: yeah, that was a mistake - there was a rethrow at the call site anyway.  Poor refactoring.

aug24:
image

Various metrics confirmed from local run.

kelvin-chappell (approved):
Happy to approve from my point of view 👍

PR #855, opened 2026-05-05
chore: Enable logging to get dev policy cache fill timings

kelvin-chappell:
This change logs a message at begin of and during fetch of dev policies so that we can see how long the process takes.

~~We're using [slow start mode](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/edit-target-group-attributes.html#slow-start-mode) for deployment so that the cache will be full when the app is ready to start receiving requests.
This log info will tell us how long to delay.~~
The above turns out to be a misunderstanding of how slow start works.  It's actually a ramping up of service capacity to allow a service to adapt itself to the shape of traffic.  Janus has no need for this so we're removing it. 
The value of this change is to see that fetching is working and roughly how long it takes.

There is already debug logging in place for individual account fetches but this extra logging is at info level to give us two extra log lines every minute, which doesn't seem excessive.  It might be useful in general to see that the service is running as expected.


app/services/DeveloperPolicyCachingService.scala
```diff
@@ -98,10 +98,26 @@ class DeveloperPolicyCachingService(
...
         // then periodically
         .append(Stream.awakeEvery[IO](fetchRate))
-        .flatMap(_ => Stream.emits(accountIams.toList))
-        .evalMap((account, iam) =>
-          fetchFromAccount(account, iam)
-            .map(status => cache.update(account, status))
+        .flatMap(_ =>
+          Stream
+            .eval(
+              logger.info("Fetching developer policies from all accounts...")
```
adamnfish-gu: Is it worth including some kind of random key here and in the log message below, to help connect these together? This would cater to the (very unlikely) situation where reloads start overlapping, but it also helps in a practical sense where you have found one half of these log lines and need to search accurately for the other.

Here's a tweak to what you have that would include this:

```scala
        ...
      ) ++ Stream
        // do first fetch immediately
        .eval(IO.randomUUID)
        // then periodically
        .append {
          Stream
            .awakeEvery[IO](fetchRate)
            .evalMap(_ => IO.randomUUID)
        }
        .flatMap { uuid =>
          Stream
            .eval(
              logger.info(
                s"${uuid.toString}: Fetching developer policies from all accounts..."
              )
            )
            .append(
              Stream
                .emits(accountIams.toList)
                .evalMap((account, iam) =>
                  fetchFromAccount(account, iam)
                    .map(status => cache.update(account, status))
                )
            )
            .append(
              Stream.eval(
                logger.info(
                  s"${uuid.toString}: Fetch of developer policies complete for all accounts."
                )
              )
            )
        }
        .handleErrorWith { err =>
          ...

```

I do wonder whether this needs to be a triply nested Stream though?

One inconsistency that jumps out is that we are mixing `Stream.append` and `++` at our different levels?

adamnfish-gu: **EDIT: No, I think this reduces our failure isolation so probably not a good idea**

The triply-nested Stream constructor is getting a little unwieldy. With these log effects around our fetch operation, it's clear that we are treating this as a single operation with a result that gets handled (logged). This directs us towards treating the fetch operation as an IO, and unnesting the Streams:

```scala
  def startPolling(): Stream[IO, Unit] =
    if (fetchEnabled) {
      val pollEvents = Stream.eval(IO.randomUUID) ++ Stream
        .awakeEvery[IO](fetchRate)
        .evalMap(_ => IO.randomUUID)
      
      Stream.eval(
        logger.info(
          s"Developer policy caching enabled with fetch rate: $fetchRate"
        )
      ) ++
        pollEvents
          .evalMap { uuid =>
            logger.info(
              s"$uuid: Fetching developer policies from all accounts..."
            ) >> accountIams.toList.traverse { (account, iam) =>
              fetchFromAccount(account, iam)
                .map(status => cache.update(account, status))
            } >> logger.info(
              s"$uuid: Fetch of developer policies complete for all accounts."
            )
          }
          .handleErrorWith { err =>
            Stream.eval(
              logger.error(err)("Failed to refresh developer policy cache")
            )
          }
    } else
      Stream.eval(logger.warn("Developer policy caching has been disabled!"))
```

See what you think!

adamnfish-gu: On second thoughts, I think that refactor is a bit less robust to failure. `IO.traverse` will stop on the first failure, which isn't what we want.

kelvin-chappell:
It's probably better to abstract out the surrounding logging but I think I'm going to close this anyway as the motivation for it is no longer relevant.  We've realised that slow starts aren't doing what we thought; they allow a slow build-up of request handling rather than a delayed start.

kelvin-chappell:
After discussion, changes coming up are:
1. Create a UUID and log it with a message at the start of every complete fetch
2. Log a single info entry per account fetch including the same UUID

This will simplify the change while allowing us to calculate roughly how long each complete fetch takes.

conf/logback.xml
```diff
@@ -38,7 +38,7 @@
     <logger name="com.amazonaws" level="WARN" />
 
     <!-- Useful for debugging role-fetching problems -->
-    <!-- logger name="services.ProvisionedRoleCachingService" level="DEBUG" /-->
+    <!-- logger name="services.DeveloperPolicyCachingService" level="DEBUG" /-->
```
adamnfish-gu, line 41: Great catch

app/services/DeveloperPolicyCachingService.scala
```diff
@@ -93,21 +94,30 @@ class DeveloperPolicyCachingService(
...
-            .map(status => cache.update(account, status))
-        )
-        .handleErrorWith { err =>
-          Stream.eval(
-            logger.error(err)("Failed to refresh developer policy cache")
+        Stream.emit(())
+          // then periodically
+          ++ Stream.awakeEvery[IO](fetchRate)
+      ).flatMap(_ =>
+        val logMarker = UUID.randomUUID()
```
adamnfish-gu: This is an effectful operation, and we're using cats-effect here. Cats Effect natively includes UUID support on `IO`, because this is a common use case. Should we be using `IO.randomUUID`?

Relatedly, our event Stream is of type `Stream[Any]` because we emit `()` and then append a stream of intervals.

Perhaps we solve both problems at once by making the event stream be a stream of UUID?

```scala
val pollEvents = Stream.eval(IO.randomUUID) ++ Stream
    .awakeEvery[IO](fetchRate)
    .evalMap(_ => IO.randomUUID)
```

We can then flatMap this stream just as we do now, with the UUID provided as the argument.

```scala
pollEvents.flatMap { logMarker =>
  ...
}
```

The `pollEvents` name is here for clarity in the comments, likely you'd inline it!

kelvin-chappell: Thanks for the suggestion!  Done in b1659d48193c66857da7b2164e8a29247c5fa3f7

adamnfish-gu (approved):
This looks ok, but will want careful testing and post release observation.

I think it's the kind of thing we'll wish we had one day when we need it, but doesn't feel urgent right now. Up to you!

kelvin-chappell:
Post-merge: Policies are still being fetched in production and the logging shows that it takes about 6s in total to fetch from all accounts.

PR #860, opened 2026-05-13
Add another dimension to metrics where suitable

aug24:

Add another metric (permission type) to existing cloudwatch metrics, where appropriate.


kelvin-chappell (approved):
👍

PR #864, opened 2026-05-19
Add dev container support

aug24:

Bring devenv config and devcontainer support up to date; add a short doc for dealing with current issues.

### DIFF
--- a/.devcontainer/devenv.yaml
+++ b/.devcontainer/devenv.yaml
@@ -21,8 +21,8 @@ modules:
 
 # Optional: Ports to forward
 forwardPorts:
-  - 9100 # same port on host and container
-  - 1234 # same port on host and container
+  - 9100 # Play app (actual actions)
+  - 1234 # Frontend app (assets for web)
 
 # Optional: Mount directories from host to container
 mounts:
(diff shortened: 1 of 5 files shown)


.devcontainer/devenv.yaml
```diff
@@ -0,0 +1,44 @@
...
+# image: "mcr.microsoft.com/devcontainers/base:ubuntu"
+
+# Optional: Ports to forward
+forwardPorts:
+  - 9100 # Play app (actual actions)
+  - 1234 # Frontend app (assets for web)
+
+# Optional: Mount directories from host to container
+mounts:
+  - "source=${localEnv:HOME}/.gu/janus-app,target=/home/vscode/.gu/janus-app,readonly,type=bind,consistency=cached"
```
adamnfish-gu, line 27: readonly here might mean this fails when the audit log DB needs updating?

adamnfish-gu (commented):
Would the docker issues be fixed by us bumping to version 3 of docker-in-docker?

https://github.com/devcontainers/features/tree/main/src/docker-in-docker

.devcontainer/devenv.yaml
adamnfish-gu: I think I might be mixing up a few things!

Could we update the docker compose file so we have a consistent definition for local dev (bare / isolated)?
https://github.com/guardian/janus-app/blob/main/local-dev/docker-compose.yml#L10

kelvin-chappell (approved):
👍

PR #868, opened 2026-05-21
Add permission.label as a metric dimension

aug24:
The permission id was used as a dimension, but it's not possible to parse dimensions with "like" when analysing data.  In order to examine the use of dev/readonly/other over time, we need the permission.label.


kelvin-chappell (approved):
👍

PR #874, opened 2026-06-05
chore: Bump devenv

kelvin-chappell:
Bumps `devenv` to the latest version.

It also fixes a couple of problems encountered when running the app in a dev container:
* Adds `node` as a dependency because it's required by frontend
* Adds a `parcel` dependency that the dev container seems to need to run the app, which is optional so as not to break the app when running locally


tjsilver (approved):
👍

PR #875, opened 2026-06-05
feat: Make passkeys mandatory

kelvin-chappell:
Currently users can opt in to step-up authentication with passkeys if they want to trial it.
In this change, we make the extra auth mandatory.
This is done by removing all references to the cookie that controls the trial.
There is still a global switch in the app config so that the feature can easily be disabled if it causes some significant unforeseen problem.

If a user's machine is compromised there is less chance that credentials will be exfiltrated.

## See also
* [Asana ticket](https://app.asana.com/1/1210045093164357/project/1213783257830028/task/1215454015085362?focus=true)

## TODO before moving out of draft state
- [ ] Comms sent around
- [ ] Agreed switch on date has arrived

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -113,14 +113,11 @@ class AppComponents(context: ApplicationLoader.Context)
         if !enabled then
           logger.warn("Passkey authentication is globally disabled!")
       )
-  private val passkeysEnablingCookieName: String =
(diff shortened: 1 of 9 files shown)


kelvin-chappell:
Superseded by #885 
Making passkeys mandatory is now just a one-line config change.

PR #881, opened 2026-06-09
feat(passkey): Clean up auth interstitial

kelvin-chappell:
This makes the auth interstitial give more of a hint that there is a process going on that's driven by browser prompts.

Screen reader just reads the dialog content.

Screenshot 2026-06-09 at 09 04 26


adamnfish-gu (approved):
This is a big improvement, thanks.

I think there are a few other things we could do to help orient users, but we can get to those separately.

PR #884, opened 2026-06-12
Add "MFA coming soon" banner

aug24:

We intend to require the use of an MFA artifact (eg hardward key, fingerprint) soon.  This change will allow us to trail the requirement ahead of time, and nudge users to ensure they have their MFA approach arranged in good time.

If no date is specified in config, or if the user has already set up MFA, no change.

Assuming the following line is present in janus config _and the user does not already have MFA_:
```
# Date we will require mfa turned on in yyyy-MM-dd format
mfaRequiredDate=2026-08-01
```

If the specified date is in the future, a banner is displayed on the main page:

image

When the specified date is in the past, the user is immediately directed to the user account page, where the message indicating they must set up a passkey is already displayed.

There is a slight disjoint in that we have a date on which we expect to turn MFA to Required, but no mechanism to ensure we do so.  However, this can remain a manual action.


aug24:
We had some discussion over a grey-list opt-out approach.  That might lead to tweaking of these messages' text and when and if they are shown, and to whom.

adamnfish-gu (commented):
For a message this important, can we stick to just the pure facts / CTA? I don't think we need to mention a "countdown", and this shifts the important information down and right.

Can we also include a CTA in it, since we're expecting an action? Ideally a styled button that goes the the personal passkey page, but even just a link would be helpful.

Many people won't know anything about this at all, so we need to make this clear and helpful, which I think the above two points would respectively aid.

Here's an example of the sort of thing I mean for the text-only version.

> Janus will require Multi-factor auth in 50 days (Sunday 8th August). Set up MFA on [your Janus account page](https://janus.gutools.co.uk/user-account).

For a more "designed" version I'd include a custom element instead of the generic toast, and design in the CTA passkey link. The generic toast is ok for a fleeting message, but it's going to be a bit jarring to be staring at this for 50 days. There's a lot of unnecessary padding, the layout doesn't help convey the message (e.g. the CTA), and the colours don't really match the rest of the application.

Up to you all!

app/conf/Config.scala
```diff
@@ -158,6 +160,27 @@ object Config {
...
+  ): Option[LocalDate] = {
+
+    config
+      .getOptional[String]("mfaRequiredDate") match {
+      case None =>
+        logger.info("No mfa required date found")
+        None
+      case Some(dateStr) =>
+        try {
+          Some(LocalDate.parse(dateStr, mfaDateFormat))
```
kelvin-chappell, line 175: Should this be UTC time to make it easier to reason about?  If I count down in Australia I can imagine windows where the count differs.

app/controllers/Janus.scala
```diff
@@ -81,6 +81,9 @@ class Janus(
           developerPolicyService.getCacheStatus,
           developerPolicyService.fetchEnabled
         )
+
+        mfaInUse = PasskeyDB.hasPasskey(request.user).toOption.getOrElse(false)
```
kelvin-chappell, line 87: If `hasPasskey` fails this will presumably show the banner.  It might be less confusing if default is true.

app/views/index.scala.html
```diff
@@ -28,7 +31,32 @@ <h1 class="header orange-text hide">Your permissions</h1>
...
-        </div>
+          </div>
+          <div role="region" aria-label="MFA status">
+            @mfaRequiredDateMaybe match {
+                case Some(d) if !mfaInUse && LocalDate.now().isBefore(d)=> {
+                    @defining(ChronoUnit.DAYS.between(LocalDate.now(), d)) { dd =>
+                        <div class="mfa-status card-panel red lighten-4" role="status" aria-atomic="true">
+                            <p>
+                                The countdown to MultiFactorAuth has started!
+                                There are @dd days remaining.
```
kelvin-chappell: Perhaps `dd` should be an optional value calculated in the model so this logic isn't visible in the view

app/views/index.scala.html
```diff
@@ -28,7 +31,32 @@ <h1 class="header orange-text hide">Your permissions</h1>
...
+                            </p>
+                            <p>
+                                Make sure you have your key by @d.
```
kelvin-chappell: Maybe link to /user-account here with text something like: 'Please register a passkey.'
And then that page will show more info about recommended passkeys etc.

app/views/index.scala.html
```diff
@@ -28,7 +31,32 @@ <h1 class="header orange-text hide">Your permissions</h1>
...
+                case Some(d) if !mfaInUse && LocalDate.now().isAfter(d) && ChronoUnit.DAYS.between(LocalDate.now(), d) < 30 => {
+                    <div class="mfa-status card-panel red lighten-2" role="status" aria-atomic="true">
+                        <p>
+                            MultiFactorAuth is now required.
+                        </p>
+                    </div>
+                }
+                case _ => {}
+            }
+          </div>
```
kelvin-chappell: This section might not be necessary because after switchover links will automatically go to the /user-account page if the user has no passkeys.  And they'll be told there that they can't continue without a passkey with a big red danger icon.

aug24: Discussed - we think it's best to have the banner up-front, rather than an unexpected effect when clicking on a link.

app/conf/Config.scala
aug24: It implicitly _is_ UTC.  There's no TZ on a local date.

app/controllers/Janus.scala
aug24: I'd rather show the warning and end up with a support call than have people locked out on the day without having seen the warning at all.

app/conf/Config.scala
kelvin-chappell: There's a period of time when it's a different date in different places but this is irrelevant here because it's always the time in Dublin anyway so I withdraw the comment.

aug24:
> For a message this important, can we stick to just the pure facts / CTA? I don't think we need to mention a "countdown", and this shifts the important information down and right.

Done

> Can we also include a CTA in it, since we're expecting an action? Ideally a styled button that goes the the personal passkey page, but even just a link would be helpful.

Done.  I prefer links for links and buttons for (usually post) actions (eg save) so I've left it as a link.

> For a more "designed" version I'd include a custom element instead of the generic toast, and design in the CTA passkey link. The generic toast is ok for a fleeting message, but it's going to be a bit jarring to be staring at this for 50 days. There's a lot of unnecessary padding, the layout doesn't help convey the message (e.g. the CTA), and the colours don't really match the rest of the application.

The colour scheme and padding is the same css we already have for other notifications.  Hopefully people won't be looking at it for long because they will sort it out.  "Jarring" might be useful for that!

app/views/userAccount.scala.html
```diff
@@ -25,6 +25,9 @@ <h1 class="header orange-text">Account</h1>
 
         <h3 class="header orange-text">Passkeys</h3>
 
+        <div class="col s12">
+        </div>
+
```
kelvin-chappell: This is a regression.
```suggestion
```

app/views/userAccount.scala.html
```diff
@@ -33,7 +36,6 @@ <h3 class="header orange-text">Passkeys</h3>
             One should be on-device if possible (e.g. Chrome browser)
             and the other off-device (e.g. security key or phone).<br />
             On a Windows device you will need two off-device passkeys as Windows Hello is disabled by default.</p>
-        <p>Guardian users can collect a Titan security key from the IT service desk in any office.</p>
```
kelvin-chappell, line 36: This is also a regression.  This should still be in the view.

app/controllers/Janus.scala
```diff
@@ -81,17 +83,39 @@ class Janus(
...
```
kelvin-chappell, line 87: Slight optimisation: if you check the passkey mode is disabled first you don't need to go to the database.  
If you want to get this out quick this could be a follow-up.

aug24: I'd rather get this out, yes.

app/controllers/Janus.scala
```diff
@@ -81,17 +83,39 @@ class Janus(
...
+        mfaComing = !mfaInUse && mfaRequiredDateMaybe
+          .exists(d => LocalDate.now().isBefore(d))
+        mfaDaysRemaining = mfaRequiredDateMaybe
+          .map(d => ChronoUnit.DAYS.between(LocalDate.now(), d))
+          .getOrElse(0L)
+        mfaDisplayDate = mfaRequiredDateMaybe
+          .map(d =>
+            d.format(DateTimeFormatter.ofPattern("d MMMM yyyy", Locale.ENGLISH))
+          )
+          .getOrElse("today")
```
kelvin-chappell, line 98: These are only relevant if mfaInUse or passkeyNotRequired so could they move into the conditional branch below?
This could be in a follow-up as well.

app/controllers/Janus.scala
```diff
@@ -81,17 +83,39 @@ class Janus(
           developerPolicyService.getCacheStatus,
           developerPolicyService.fetchEnabled
         )
+
+        mfaInUse = PasskeyDB.hasPasskey(request.user).toOption.getOrElse(false)
+        mfaRequiredDateMaybe = Config.mfaRequiredDataMaybe(configuration)
+        mfaComing = !mfaInUse && mfaRequiredDateMaybe
```
kelvin-chappell, line 89: I think this should be false if Passkey mode is disabled

conf/application.conf
```diff
@@ -112,3 +112,5 @@ developerPolicies {
   }
   discoverablePath = "/developer-policy/"
 }
+
+mfaRequiredDate = "2026-07-01"
```
kelvin-chappell, line 116: Presumably this is overridden in Prod config?

kelvin-chappell (approved):
👍 just a couple of refactorings/optimisations for a follow-up

app/controllers/Janus.scala
aug24: Discussed: no, it should not, because showing the banner should happen anyway.  We ARE going to do this.

conf/application.conf
aug24: Yes, it is.

PR #889, opened 2026-06-17
Improve setup

aug24:

Improve the setup process by doing the test code changes to create tables, running those tests, and then reverting.


scripts/setup.sh
```diff
@@ -31,19 +31,30 @@ else
...
 fi
 
+docker compose -f local-dev/docker-compose.yml up -d
+sleep 2
+
+# Creating tables (if needed - aim for idempotency)
+aws dynamodb list-tables --profile security --region eu-west-1 --endpoint http://localhost:8000 \
+  | jq -r '.TableNames[]' \
+  | grep -w PasskeyChallenges  \
+  || sbt "Test / runMain aws.PasskeyChallengeDBTestApp create"
```
kelvin-chappell: These three commands are very similar and there's scope for errors in copying and pasting without changing the class name for example.  Could it be extracted as a function?

test/aws/AuditTrailDBTest.scala
```diff
@@ -45,7 +45,7 @@ class AuditTrailDBTest extends AnyFreeSpec with Matchers {
       println(userResults.toList)
     }
 
-    "create database table" ignore {
+    "create table" ignore {
       createTable()
     }
```
kelvin-chappell: Do we need this any more?  As it was never really a test anyway apart from I suppose a test that the table could be created.
Similar for the other 'tests' for the other tables.

scripts/setup.sh
aug24: I'd have just looped over the table names if the test file naming convention was consistent.  But I can certainly make a tiny function with two params.

test/aws/AuditTrailDBTest.scala
```diff
@@ -141,3 +141,17 @@ class AuditTrailDBTest extends AnyFreeSpec with Matchers {
     dynamoDB.deleteTable(request)
   }
 }
+
+object AuditTrailDBTestApp {
+
+  def main(args: Array[String]): Unit = {
```
adamnfish-gu: All the table names and logic is available here, so it's a bit strange to see it be duplicated into the bash script.

Could this program be a self contained "put the DB in the correct state", with its own condition checks?

It's also very unusual to see `Test / run`, but I don't feel strongly that this needs to be pulled out into its own main because that will likely mean adding a whole new subproject. The pragmatism of this looks ok to me, but maybe let's aim to make it one program with one purpose, rather than 3 with 2. We don't need destroy, and we don't need separate mains for each table. Let's have a single CLI endpoint for creating both tables?

scripts/setup.sh
```diff
@@ -31,19 +31,30 @@ else
   echo "janus-service-account-cert.json is present"
 fi
 
+docker compose -f local-dev/docker-compose.yml up -d
+sleep 2
+
+# Creating tables (if needed - aim for idempotency)
+aws dynamodb list-tables --profile security --region eu-west-1 --endpoint http://localhost:8000 \
```
adamnfish-gu: As mentioned in mire detail on the other file, this pulls more logic (and names) into the script than I'd like, which means things look a bit scattered. Can we make this a single idempotent invocation here? This would also make it more usable for people that want to run this "stand-alone".

build.sbt
```diff
@@ -169,3 +169,11 @@ lazy val configTools = (project in file("configTools"))
     name := "janus-config-tools",
     description := "Library for reading and writing Janus configuration files"
   )
+
+lazy val Setup = (project in file("setup"))
```
adamnfish-gu: This should be `setup` not `Setup`. Upper case makes it look like a scope (Test / Run / Compile etc) so it's super confusing to have this be `Setup` because the `Setup / run` looks like run in the setup scope (rather than the setup project).

setup/resources/logback-test.xml
```diff
@@ -0,0 +1,37 @@
...
+        <encoder>
+            <pattern>%date{yyyy-MM-dd HH:mm:ss ZZZZ} [%thread] %-5level %logger{36} - %msg%n%xException{3}</pattern>
+        </encoder>
+    </appender>
+
+
+    <logger name="play" level="INFO" />
+    <logger name="application" level="DEBUG" />
+
+    <!-- disable pan-domain auth's spam -->
```
adamnfish-gu: Do we need this here?

I wonder if this file has been copied from elsewhere?

test/aws/PasskeyDBTest.scala
```diff
@@ -9,69 +9,7 @@ import software.amazon.awssdk.services.dynamodb.model._
...
-      destroyTable()
-    }
-  }
-
-  /** NB: Only use these for local testing use the provided CloudFormation
-    * template to create table in AWS environments.
-    *
-    * If you update this then be sure to also update the CloudFormation
-    * template's definition.
+  /** Once the table create code was removed, this test class was empty. TODO:
```
adamnfish-gu: Why would we leave an empty test class around? If this isn't needed that's ok and we can remove it?

scripts/setup.sh
```diff
@@ -31,19 +31,17 @@ else
   echo "janus-service-account-cert.json is present"
 fi
 
+docker compose -f local-dev/docker-compose.yml up -d
```
adamnfish-gu, line 44: I think we need to bring this back down when we're done. I'm also looking anxiously at that hopeful "sleep 2", below. I don't think that'll work on your home network if you don't have the DynamoDB image cached locally.

Is there a better way to handle this?

scripts/setup.sh
```diff
@@ -31,19 +31,17 @@ else
...
+sleep 2
+
+# Creating tables (if needed - aim for idempotency)
+sbt "Setup / runMain aws.DBSetupApp"
+
 if [[ ! -f ~/.gu/janus-app/janusData.conf ]]; then
-   echo "You need to copy in the janusData file: ~/.gu/janus-app/janusData.conf"
+   echo "!!! You need to copy in a janusData.conf file at ~/.gu/janus-app/janusData.conf"
```
adamnfish-gu, line 51: Do we need to provide a bit of guidance on how to do this, if we're trying to make this easy? We don't necessarily have to do that here, but if the goal is general "improve setup" then maybe a slightly clearer message would help.

scripts/setup.sh
```diff
@@ -31,19 +31,17 @@ else
   echo "janus-service-account-cert.json is present"
 fi
 
+docker compose -f local-dev/docker-compose.yml up -d
+sleep 2
+
+# Creating tables (if needed - aim for idempotency)
+sbt "Setup / runMain aws.DBSetupApp"
```
adamnfish-gu: Given that this project is named after a single use case, can this just be `setup/run`, with a single Main program in there? That has one job at the moment (set up the DB), which is fine! It removes a piece of required knowledge from being able to use this - we're just exposing a "setup" program and it is our business what it does.

setup/resources/logback-test.xml
```diff
@@ -0,0 +1,37 @@
+<configuration>
+
+    <contextName>janus</contextName>
+
+    <appender name="LOGFILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>logs/janus-test.log</file>
```
adamnfish-gu: Is this the right log name?

setup/src/main/scala/aws/AuditTrailDBSetup.scala
```diff
@@ -0,0 +1,101 @@
...
+import logic.AuditTrail
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient
+import software.amazon.awssdk.services.dynamodb.model.KeyType.{HASH, RANGE}
+import software.amazon.awssdk.services.dynamodb.model.ScalarAttributeType.{N, S}
+import software.amazon.awssdk.services.dynamodb.model._
+
+import java.time.ZoneOffset.UTC
+import java.time.{Duration, ZonedDateTime}
+
+class AuditTrailDBSetup {
```
adamnfish-gu: Why is this a class? It looks like it only contains static methods, could this just be an object that contains functions? The class just creates the indirection of having to create an instance before we can access the methods.

The same is true of all the others as well, I think.

setup/src/main/scala/aws/DBSetupApp.scala
```diff
@@ -0,0 +1,34 @@
+package aws
+
+import play.api.Logging
+import software.amazon.awssdk.services.dynamodb.model.*
+
+object DBSetupApp extends Logging {
+
+  def main(args: Array[String]): Unit = {
+    try {
```
adamnfish-gu: Is this definitely what we want!?

I'd be pretty nervous about this setup script nuking a working DB(!)

I think checking if the table exists would be much more helpful to a working engineer than making sure that's the case by nuking whatever's there 😅

setup/src/main/scala/aws/DBSetupApp.scala
```diff
@@ -0,0 +1,34 @@
...
+          s"Passkey challenge table delete skipped with ${e.getMessage}"
+        )
+    }
+    try {
+      new PasskeyDBSetup().destroyTable()(Clients.localDb)
+    } catch {
+      case e: Throwable =>
+        logger.info(s"Passkey table delete skipped with ${e.getMessage}")
+    }
+    new AuditTrailDBSetup().createTable()(Clients.localDb)
```
adamnfish-gu: There's no error handling here. Probably fine, but it does mean this script can put the DB into an inconsistent state. We don't necessarily need production grade edge-case handling here, but I think it might be nice to try all three and report on the overall status at the end, so someone coming to this has an idea of what happened after the fact?

build.sbt
adamnfish-gu: This project needs the `publish / skip := true` setting, otherwise it'll try and publish this as a library whenever we run the publish step for configTools.

adamnfish-gu (changes requested):
Some questions and a few important points in here. One super important thing is skipping the publish step for the new sbt project! I'd also expect to see this exercised in CI, even if it's just to compile it?

I'm **super nervous** about how it drops the DBs when it is run, now 👀 

I'm also pretty sure there are comments here and/or in the private Janus repo's infrastructure that say "make sure these DB definitions are kept in sync". We'll need to hunt for those to make sure they're all still accurate.

build.sbt
aug24: Done

setup/src/main/scala/aws/DBSetupApp.scala
aug24: Done.  It'll try to delete all three (if requested) and then report, then try to create all three (if requested) and report.

setup/src/main/scala/aws/DBSetupApp.scala
aug24: I've made it so the user has to explicitly request what they want: destroy and/or create.

setup/src/main/scala/aws/AuditTrailDBSetup.scala
aug24: Done.

setup/resources/logback-test.xml
aug24: I should have deleted a big chunk of it and altered the log output.  Done.

scripts/setup.sh
aug24: Finding and pulling images blocks, only container running goes to daemon.  That 2s is just to allow the container to initialise itself.

scripts/setup.sh
aug24: I wasn't sure, as this is a public repo, if we wanted to document that here other than to say you need it.

test/aws/PasskeyDBTest.scala
aug24: Kelvin confirms there are no tests to put in.  I'll delete.

build.sbt
aug24: Done

scripts/setup.sh
aug24: Agreed.  Done.

setup/resources/logback-test.xml
aug24: That file needed substantial fixing. Doesn't need any of the play stuff either.  Done.

aug24:
> Some questions and a few important points in here. One super important thing is skipping the publish step for the new sbt project!

Done.

> I'd also expect to see this exercised in CI, even if it's just to compile it?

Done.  I don't think tiny internal tools like this really deserve anything stronger.

> I'm **super nervous** about how it drops the DBs when it is run, now 👀

I was aiming for an idempotent "set up clean tables" approach, but I've made it more explicit instead.

> I'm also pretty sure there are comments here and/or in the private Janus repo's infrastructure that say "make sure these DB definitions are kept in sync". We'll need to hunt for those to make sure they're all still accurate.

There are.  I can check if you like, but these are the tables we've been using for development for the passkey work, for instance, so they're pretty fresh, and work.

.github/workflows/trigger-private-janus-build.yml
```diff
@@ -50,11 +50,12 @@ jobs:
...
 
       - run: >
-          sbt 
-          compile 
-          Test/compile 
-          scalafmtCheckAll 
-          scalafmtSbtCheck 
+          sbt
+          compile
+          Test/compile
```
adamnfish-gu, line 55: This `Test/compile` followed by `setup/compile` is a mildly odd, but it's not a huge deal here in CI.

If we add the setup project to the root aggregation then it'll get done automatically by `Test/compile`. `Test/compile` says run compile on the tests, and of course that will also mean the code itself gets compiled (since the tests depend on the code). Aggregation says "when a task is run here, also run it on these other projects", and is commonly used to set up the "root" project to run things everywhere.

It's normal for root to aggregate all the other projects, and if that were done we wouldn't need the explicit `setup/compile` line.

scripts/setup.sh
adamnfish-gu: Yeah no harm, perhaps we can think about that later.

scripts/setup.sh
```diff
@@ -47,12 +40,17 @@ aws --profile security --region eu-west-1 ssm get-parameter \
...
 
-sbt "testOnly *DBTest*"
+# Create tables - will fail if they exist, but can be run manually with "destroy" or "recreate"
+sbt "setup / run create"
+
+if [[ ! -f ~/.gu/janus-app/janusData.conf ]]; then
+   echo "!!! You need to copy in a janusData.conf file at ~/.gu/janus-app/janusData.conf"
+   exit 1
+else
+  echo "janusData.conf is present"
```
adamnfish-gu, line 54: Do we need to say this? No dramas, but it's nice to recuce noise.

setup/src/main/scala/aws/DBSetupApp.scala
```diff
@@ -0,0 +1,128 @@
...
+          logger.error(s"Passkey table create failed: ${e.getMessage}")
+          false
+      }
+      if (!passkey || !passkeyChallenge || !audit) {
+        System.exit(1)
+      }
+    }
+  }
+
+  private def parseArgs(args: List[String]): (Boolean, Boolean) = {
```
adamnfish-gu: Is it worth adding a couple of tests for this? It would document the expected behaviour and prevent regressions, and this is mildly fiddly.

adamnfish-gu (approved):
Couple of little points but it's basically fine - and a big improvement, thanks!

scripts/setup.sh
aug24: Each of the previous steps also confirms that the check has passed.  I think I'll leave it, at least for now.

setup/src/main/scala/aws/DBSetupApp.scala
aug24: Done

.github/workflows/trigger-private-janus-build.yml
aug24: If we add setup to root's aggregate, it can't depend on root, which is where the clientdb object lives.  I'd have to refactor a bit.  Will investigate...

aug24: This isn't worth it!

kelvin-chappell (approved):

PR #890, opened 2026-06-17
feat(setup): Automate AWS profile config

kelvin-chappell:
## Problem
We currently have quite an awkward manual process for getting hold of 'janus' AWS profile credentials, which have to be permanent for reasons explained [here]().
This manual process doesn't work in a repeatable dev container at all.

## Solution implemented
We now store the creds in an encrypted SSM param and install them locally as part of the new scripted setup process.

## Tested
Successfully ran the setup script using the standard dev permission locally and in a dev container in VSCode.
Verified that the janus profile ends up in the dev container's local creds store.
Running the app using the start script in a dev container allows getting creds for the dev playground account.

## See also
* The dev policy for running Janus locally needs to be able to read the new param:
https://github.com/guardian/janus/pull/5574

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -4,7 +4,7 @@ set -e
 
 # Create a placeholder directory for janus-app config in ~/.gu
 mkdir -p ~/.gu/janus-app/data
-chown -R $(whoami):$(whoami) ~/.gu/janus-app/data
+chown -R $(whoami):staff ~/.gu/janus-app/data
 
 aws --profile security --region eu-west-1 sts get-caller-identity >/dev/null
 if [[ $? -ne 0 ]]; then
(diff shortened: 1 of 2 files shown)


scripts/setup.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 # Create a placeholder directory for janus-app config in ~/.gu
 mkdir -p ~/.gu/janus-app/data
-chown -R $(whoami):$(whoami) ~/.gu/janus-app/data
+chown -R $(whoami):staff ~/.gu/janus-app/data
```
adamnfish-gu, line 7: Is this because of the mount?

scripts/setup.sh
```diff
@@ -38,6 +38,15 @@ else
   echo "janusData.conf is present"
 fi
 
+# Need AWS profile 'janus' to test access to dev-playground account.
```
adamnfish-gu, line 41: Should this avoid over-writing an existing credential if it exists? I'm not sure, maybe it is best to have this create a consistent environment!

adamnfish-gu (approved):
Couple of questions for my own info, but looks like a great idea.

scripts/setup.sh
kelvin-chappell: It was a pragmatic solution to the script not working locally.  It works in the dev container because there's a group called vscode but it wouldn't work generally.  Do we actually need to change the group at all?

scripts/setup.sh
kelvin-chappell: If we rotate the key it's probably going to be a smoother experience to just be given the new creds.  I can't think of an obvious scenario in which you'd want to keep different creds.  But we can always change it later if so.

scripts/setup.sh
aug24: FYI We took the mount out altogether in favour of fetching (with run locally credentials) files to produce a fully configured dev container.

PR #898, opened 2026-06-23
Fix odd build errors

aug24:

We have build errors on these two lines.


aug24:
Merged into the patch that needed it.

PR #899, opened 2026-06-25
Refactor for clarity

aug24:
This is a follow on change from the MFA Banner to make the code paths clearer.

### DIFF
--- a/app/controllers/Janus.scala
+++ b/app/controllers/Janus.scala
@@ -85,22 +85,25 @@ class Janus(
         )
 
         mfaInUse = PasskeyDB.hasPasskey(request.user).toOption.getOrElse(false)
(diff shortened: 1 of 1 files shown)


app/controllers/Janus.scala
```diff
@@ -84,23 +84,26 @@ class Janus(
...
+            .getOrElse(0L)
+          val mfaDisplayDate = mfaRequiredDateMaybe
+            .map(d =>
+              d.format(
+                DateTimeFormatter.ofPattern("d MMMM yyyy", Locale.ENGLISH)
+              )
+            )
+            .getOrElse("today")
+          val passkeyNotRequired = Config.passkeyMode(
+            configuration
```
kelvin-chappell: Isn't this needed for the condition above?

kelvin-chappell (approved):
👍

PR #900, opened 2026-06-25
Handle mixed case email addresses in passkey checks

aug24:

Ensure that the email address used to check passkeys is also lower cased.

This is the minimal version to solve the problem of "Justin.Rowles" vs "justin.rowles".  The maximal version is in https://github.com/guardian/janus-app/pull/901, which has the same principle applied to everywhere that passes in a username.

### DIFF
--- a/app/controllers/PasskeyController.scala
+++ b/app/controllers/PasskeyController.scala
@@ -271,11 +271,12 @@ class PasskeyController(
 
   private def verifyHasAccess(
       user: UserIdentity
-  ): Try[Unit] =
-    if hasAccess(user.username, janusData.access) ||
-      hasAccess(user.username, janusData.admin)
+  ): Try[Unit] = {
+    if hasAccess(user, janusData.access) ||
+      hasAccess(user, janusData.admin)
     then Success(())
     else Failure(JanusException.noAccessFailure(user))
+  }
 
   private def loadRegistrationChallenge(
       user: UserIdentity
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -38,7 +38,7 @@
                         <ul class="right hide-on-med-and-down">
                             <li><a href="/accounts">Accounts</a></li>
                             <li><a href="/support">Support</a></li>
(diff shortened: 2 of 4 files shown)


app/controllers/PasskeyController.scala
```diff
@@ -271,11 +271,12 @@ class PasskeyController(
 
   private def verifyHasAccess(
       user: UserIdentity
-  ): Try[Unit] =
-    if hasAccess(user.username, janusData.access) ||
-      hasAccess(user.username, janusData.admin)
+  ): Try[Unit] = {
+    if hasAccess(user, janusData.access) ||
+      hasAccess(user, janusData.admin)
```
aug24, line 276: Originally I was going to put the `username(user)` call here to replace `user.username` but it makes it much more of a pain to write the test.

test/logic/UserAccessTest.scala
```diff
@@ -533,12 +533,22 @@ class UserAccessTest
...
     )
 
+    def getUserIdentityForEmail(email: String) =
+      UserIdentity("", email, "", "", 0L, None)
+
     "returns true when given a user that has an entry" in {
-      hasAccess("test.user", adminACL) shouldEqual true
+      val user = getUserIdentityForEmail("test.user@guardian.co.uk")
+      hasAccess(user, adminACL) shouldEqual true
+    }
```
aug24, line 542: In the "further work" version of this PR, this approach is extended substantially, leaving almost all the code references to username(user) in one file.

app/views/main.scala.html
```diff
@@ -47,7 +47,7 @@
                             <li><a href="/">Home</a></li>
                             <li><a href="/accounts">Accounts</a></li>
                             @if(isSupportUser(username(user), Instant.now(), janusData.support)) { <li><a href="/support">Support</a></li> }
```
aug24, line 49: Note that this line has `username(user)`, but there's no test for mixed case for isSupportUser, because it's not easy to test controllers or html files.

kelvin-chappell:
I much prefer #901

PR #901, opened 2026-06-25
Handle mixed case email addresses in general

aug24:
We have, so far, converted a Google UserIdentity to an email local section at the call site.  This may be simpler if it's done at the usage site.  Discuss.

### DIFF
--- a/app/controllers/Janus.scala
+++ b/app/controllers/Janus.scala
@@ -117,8 +117,7 @@ class Janus(
                 mfaDisplayDate
               )
           )
-        else
-          SeeOther(controllers.routes.Janus.userAccount.url)
+        else SeeOther(controllers.routes.Janus.userAccount.url)
       }) getOrElse Ok(views.html.noPermissions(request.user, janusData))
     }
 
--- a/app/controllers/Janus.scala
+++ b/app/controllers/Janus.scala
@@ -64,12 +64,12 @@ class Janus(
         Date.displayMode(ZonedDateTime.now(ZoneId.of("Europe/London")))
       (for {
         accountsAccess <- internalUserAccess(
-          username(request.user),
+          request.user,
           janusData,
           developerPolicyService.getDeveloperPolicies
         )
         userPolicyGrants = policyGrantsForUser(
-          username(request.user),
+          request.user,
           janusData.access
         )
         favourites = Favourites.fromCookie(request.cookies.get("favourites"))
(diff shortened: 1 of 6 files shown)


app/controllers/Janus.scala
```diff
@@ -64,12 +64,12 @@ class Janus(
         Date.displayMode(ZonedDateTime.now(ZoneId.of("Europe/London")))
       (for {
         accountsAccess <- internalUserAccess(
-          username(request.user),
+          request.user,
```
aug24, line 67: Getting the username at the call site must always be done (in multiple places) with `username(user)` not `user.username`.  Both give plain old strings, which is unhelpful.  This is open to error, obviously, so just pass it in and the usage site do it once.

12 call site usages turn to five usage site usages.

app/controllers/Janus.scala
```diff
@@ -64,12 +64,12 @@ class Janus(
         Date.displayMode(ZonedDateTime.now(ZoneId.of("Europe/London")))
       (for {
         accountsAccess <- internalUserAccess(
-          username(request.user),
```
aug24, line 67: First of twelve of these (including the ones missing from hasAccess)

test/logic/UserAccessTest.scala
```diff
@@ -24,6 +24,9 @@ class UserAccessTest
 
   import UserAccess.*
 
+  def getUserIdentityForEmail(email: String) =
```
kelvin-chappell: [Nitpick] In case future readers get confused, it's really the username rather than the email, as UserIdentity has an `email` field, which is `a.b@c.d`, and a derived `username` field, which is the one we're using:
```suggestion
  def getUserIdentityFromUsername(username: String) =
```
or
```suggestion
  def toUser(username: String) =
```

test/logic/UserAccessTest.scala
```diff
@@ -534,11 +545,18 @@ class UserAccessTest
     )
 
     "returns true when given a user that has an entry" in {
-      hasAccess("test.user", adminACL) shouldEqual true
+      val user = getUserIdentityForEmail("test.user")
+      hasAccess(user, adminACL) shouldEqual true
+    }
+
+    "returns true when given a user that has an entry but Google Auth has given us a capitalised email" in {
```
kelvin-chappell:
```suggestion
    "returns true when given a user that has an entry but Google Auth has given us a capitalised username" in {
```

kelvin-chappell (approved):
👍

PR #902, opened 2026-06-25
css to control spinner rotation direction

andrew-nowak:

Updates the spinner on the passkey auth page to rotate in the same direction as the arrows point

Before:

https://github.com/user-attachments/assets/830b5b74-b8fa-4b1f-8a3e-c18e4c489d03

After:

https://github.com/user-attachments/assets/766a4afa-f854-4ea9-a080-418208c37db5

Andrew gets less distracted by the spinner on the passkey auth page 😅


tjsilver (approved):
Now I'm thinking - why aren't the arrows pointing in a clockwise direction?!

kelvin-chappell:
I only wish I had this attention to detail!

PR #903, opened 2026-06-26
Provide DEV configuration file path to sbt

adamnfish-gu:

When running the Play app locally using `sbt`, we now provide the path of the local configuration file. This prevents the user needing to remember to pass the correct `-Dconfig.file` property to sbt on every invocation.

This makes it easier to set up a local dev environment, and simpler to use.


kelvin-chappell (approved):
👍

PR #904, opened 2026-06-29
Improve passkey banner display

adamnfish-gu:

The passkey banner is front and centre for all our users (now only most, if some have set up MFA). This means it warrants a bit of design attention.

This PR introduces natively styled banners, and uses them for both the MFA notification and the existing developer policy service status messages.

Relatedly, I've switched the MFA banner from "error" to "warning". I think this is correct, because until MFA is required it isn't an error state to not have MFA. If we add a "you cannot proceed without MFA" banner in the future, I think that one **would** be an error.

This takes up a lot less screen space at the top of the critical permissions page, and the design looks much more integrated in to Janus.

#### Error banner
Screenshot 2026-06-29 at 10 32 32

#### Warning banner
Screenshot 2026-06-29 at 10 32 43

#### Info banner
Screenshot 2026-06-29 at 10 32 51

#### Stacked banners
Screenshot 2026-06-29 at 10 33 48

### DIFF
--- a/app/logic/ViewHelpers.scala
+++ b/app/logic/ViewHelpers.scala
@@ -56,4 +56,17 @@ object ViewHelpers {
     case MaterialIcon(name: String)
     case AssetImage(path: String, alt: String)
   }
+
+  enum CtaLevel {
+    case Warn
+    case Error
+    case Info
+  }
+  object CtaLevel {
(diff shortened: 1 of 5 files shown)


tjsilver (approved):
Looks good!

PR #908, opened 2026-07-07
Improve account page passkey docs

adamnfish-gu:

Improves the readability of the passkey instructions on the account page, by adding some design elements. This lays the content out in a way that helps the reader understand the context of each part of our recommendations.

I think this improved design gives us the cognitive headroom to restore previously removed content (about Firefox vs Chrome), which should help address one of the common FAQ items for the passkey rollout.

Even after passkeys are required, this screen will be important to new users. It'll help to be as clear as possible about what we expect users to do.

### New design

Screenshot 2026-07-07 at 12 02 27

### The existing version, for reference

Screenshot 2026-07-07 at 12 02 42


kelvin-chappell (approved):
🚀

adamnfish-gu:
We'll separately look at the use of `grey-text`, because it may not have enough contrast. We use it here to be consistent with other areas (like developer-policies), but perhaps we need to use something with a little more contrast.

PR #909, opened 2026-07-07
Improve readability of passkey docs information

adamnfish-gu:

We shouldn't be using light grey text for prose, because the contrast isn't enough to make it readable. The light grey is ok for design elements (as used elsewhere in the app), but here we need to darken it.

Improves the accessibility of this information, while preserving the visual structure that the previous design introduced.

### Tweaked design
Screenshot 2026-07-07 at 16 02 09

### The existing version, for reference
Screenshot 2026-07-07 at 12 02 27


kelvin-chappell (approved):
👍

PR #922, opened 2026-07-20
Provide clarity to users for "admin" permissions

adamnfish-gu:

**Direct Janus' users away from sensitive permissions outside of limited administrative activities.**

Uses the existing `shortTerm` flag on permissions to underpin modelling and UX for admin permissions. This allows us to provide a UX nudge away from using those permissions for day-to-day work to reduce the risk of supply-chain attacks and AI tooling.

We'll expect to see reduced use of administrative permissions, especially for CLI sessions via Janus-bestowed credentials.

## Images

Administrative permissions are grouped in their own section, which includes a warning notice at the top. Here's how it appears on the UI.

Screenshot 2026-07-20 at 10 22 22

## Any additional notes?

This will come with some tweaks to our permissions in the config repository as well.

It's worth calling out that because re-uses the `shortTerm` flag, we only get this behaviour for permissions that can tolerate the lower session times. This makes sense because it prevents the multi-step rollout of a config change, and means we don't have two mostly overlapping concepts in the config format and codebase(s).

### DIFF
--- a/app/logic/AccountOrdering.scala
+++ b/app/logic/AccountOrdering.scala
@@ -28,9 +28,12 @@ object AccountOrdering {
         else favIndex
       }
       .map { case (account, accountAccess) =>
+        val (adminPermissions, standardPermissions) =
+          partitionAdminPermissions(accountAccess.permissions.sorted)
         AwsAccountAccess(
           awsAccount = account,
-          permissions = accountAccess.permissions.sorted,
+          standardPermissions = standardPermissions,
+          adminPermissions = adminPermissions,
           developerPolicies = accountAccess.developerPolicies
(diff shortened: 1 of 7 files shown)


kelvin-chappell (commented):
Does this make the '/superuser' route confusing?  We now have two sets of admin permissions

kelvin-chappell (approved):
👍 Just a comment for clarification

adamnfish-gu:
> Does this make the '/superuser' route confusing? We now have two sets of admin permissions

This is a great point. I think that will need a follow up. We don't have any non-P&E admin users that would be affected by this at the moment, so I think it's ok to do this separately. What do you think?

PR #925, opened 2026-07-20
Extends the maximum session length for "short-term" permissions

adamnfish-gu:

Extends the maximum time for `shortTerm` permissions, and provides more flexibility for short term session length selection in the UI.

This may be required as a temporary measure while if we mark more permissions as being short-term. Let's be driven by the need, if it arises.

See also: https://github.com/guardian/janus-app/pull/922, which makes it clearer to users of Janus which of their permissions should be considered "admin".

## Images

The select still defaults to 1 hour:
Screenshot 2026-07-20 at 17 50 40

But now includes more (and longer) options:
Screenshot 2026-07-20 at 17 50 28

### DIFF
--- a/app/aws/Federation.scala
+++ b/app/aws/Federation.scala
@@ -24,7 +24,7 @@ object Federation {
   /** Credential/Console lease times. Defaults are used when user doesn't
     * request specific time periods, max will limit how long can be requested.
     */
-  val maxShortTime: Duration = 3.hours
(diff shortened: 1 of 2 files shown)


adamnfish-gu:
We're a few weeks into the session length reduction, and haven't needed this yet. I'll close this PR, it's still here if we ever need to go back to it.

PR #931, opened 2026-07-29
Rename AccessSource.Admin to AccessSource.Superuser

adamnfish-gu:
> [!NOTE]
> **Disclaimer**: AI was used to significantly generate all the stacked PRs associated with this change. I've manually reviewed all the work and taken time to edit and curate the AI's output (both code and text), but this work has primarily been done by Opus 5 via the GitHub Copilot CLI.

Renames `AccessSource.Admin` to `AccessSource.Superuser`, continuing the rename from #930.

> [!WARNING]
> **This changes an emitted CloudWatch metric dimension.** The `access-source` dimension will start emitting `Superuser` instead of `Admin`, which starts a new time series. Existing dashboards and alarms that filter on `access-source = Admin` will need updating.

The dimension value comes from the enum name, so it changes with the rename.

---

Part of #937, which tracks the whole migration including the releases,
deploys and verification steps between these PRs.

1. app: rename estate-wide concept to superuser (#930)
1. app: rename `AccessSource.Admin` to `Superuser` (#931)  ← **you are here**
1. configTools: read both `superuser` and legacy `admin` keys (#932)
1. guardian/janus and other consumers: bump to 11.1.0
1. configTools: write `superuser`, rename `JanusData.admin` (#933)
1. example project pinned to 12.0.0 (#934) — CI red until 12.0.0 ships
1. guardian/janus and other consumers: bump to 12.0.0
1. configTools: remove legacy `admin` read support (#935)
1. guardian/janus and other consumers: bump to 13.0.0

PR #930, opened 2026-07-29
Rename the estate-wide "admin" concept to "superuser"

adamnfish-gu:
> [!NOTE]
> AI was used to significantly generate all the stacked PRs associated with this change. I've manually reviewed all the work and taken time to edit and curate the AI's output (both code and text), but this work has primarily been done by Opus 5 via the GitHub Copilot CLI.

Renames the historic concept of "admin" (estate-wide oversight access) to "superuser" throughout the application.

Janus now has two concepts competing for the word "admin":

- **short-term permissions**, surfaced to users as "admin" permissions to steer them away from day-to-day use
- **estate-wide oversight access**, configured in a separate ACL and shown on its own page

This duplication was introduced in https://github.com/guardian/janus-app/pull/922, becuase we now want to be able to signpost powerful permissions as "admin" to discourage day-to-day use. The older "estate-wide oversight" use of admin is very limited, so that concept doesn't need to hold onto the name.

The oversight route [was already renamed](https://github.com/guardian/janus-app/pull/438) to `/superuser` for infra reasons, which gives us a clue how to proceed. This stack of PRs adopts the name "superuser" for oversight access, across the rest of the app.

In this PR we start the renaming with the controller action, the view, `UserAccess` functions and parameters, `SourcedAccountAccess.admin`, the nav labels, the CSS classes and the example project's `Admin.scala`.

The `janus-config-tools` config key is not touched here, so the data format for configuration files remains unchanged. This stack of PRs takes us through the full migration, which will include making two major version bumps to the configTools library so we can do a multi-step rollout without breaking changes.

---

Part of #937, which tracks the whole migration including the releases,
deploys and verification steps between these PRs.

1. app: rename estate-wide concept to superuser (#930)  ← **you are here**
1. app: rename `AccessSource.Admin` to `Superuser` (#931)
1. configTools: read both `superuser` and legacy `admin` keys (#932)
1. guardian/janus and other consumers: bump to 11.1.0
1. configTools: write `superuser`, rename `JanusData.admin` (#933)
1. example project pinned to 12.0.0 (#934) — CI red until 12.0.0 ships
1. guardian/janus and other consumers: bump to 12.0.0
1. configTools: remove legacy `admin` read support (#935)
1. guardian/janus and other consumers: bump to 13.0.0

PR #932, opened 2026-07-29
configTools: read the superuser ACL from either config key

adamnfish-gu:
> [!NOTE]
> **Disclaimer**: AI was used to significantly generate all the stacked PRs associated with this change. I've manually reviewed all the work and taken time to edit and curate the AI's output (both code and text), but this work has primarily been done by Opus 5 via the GitHub Copilot CLI.

Teaches the config loader to read the superuser ACL from either `janus.superuser` or the legacy `janus.admin` key.

This is the first step of the config format migration, which is split up to allow us to roll back. This change means the Janus application understands both formats, so a config file in either shape is valid.

- `Loader.loadAdmin` becomes `loadSuperuser`, reading `janus.superuser` first and falling back to `janus.admin`
- `ConfiguredAdmin` becomes `ConfiguredSuperuser`
- If a config file contains both keys, `superuser` wins
- If neither key is present, loading fails and the error names the new key, `janus.superuser`

The writer, the Twirl template and `JanusData.admin` are unchanged, so this is not a breaking change and no release is required.

---

Part of #937, which tracks the whole migration including the releases,
deploys and verification steps between these PRs.

1. app: rename estate-wide concept to superuser (#930)
1. app: rename `AccessSource.Admin` to `Superuser` (#931)
1. configTools: read both `superuser` and legacy `admin` keys (#932)  ← **you are here**
1. guardian/janus and other consumers: bump to 11.1.0
1. configTools: write `superuser`, rename `JanusData.admin` (#933)
1. example project pinned to 12.0.0 (#934) — CI red until 12.0.0 ships
1. guardian/janus and other consumers: bump to 12.0.0
1. configTools: remove legacy `admin` read support (#935)
1. guardian/janus and other consumers: bump to 13.0.0

### DIFF
--- a/configTools/src/main/scala/com/gu/janus/config/Loader.scala
+++ b/configTools/src/main/scala/com/gu/janus/config/Loader.scala
@@ -13,15 +13,18 @@ import scala.util.Try
   * fails, a description of the failure is made available.
   */
 object Loader {
+  private val superuserPath = "janus.superuser"
+  private val legacySuperuserPath = "janus.admin"
+
   def fromConfig(config: Config): Either[String, JanusData] = {
     for {
       permissionsRepo <- loadPermissionsRepo(config)
       accounts <- loadAccounts(config)
       permissions <- loadPermissions(config, accounts)
       access <- loadAccess(config, permissions)
-      admin <- loadAdmin(config, permissions)
+      superuser <- loadSuperuser(config, permissions)
       support <- loadSupport(config, permissions)
-    } yield JanusData(accounts, access, admin, support, permissionsRepo)
+    } yield JanusData(accounts, access, superuser, support, permissionsRepo)
   }
 
   private[config] def loadPermissionsRepo(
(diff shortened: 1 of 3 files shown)


configTools/src/main/scala/com/gu/janus/config/Loader.scala
```diff
@@ -114,22 +117,32 @@ object Loader {
...
+    * when both keys are present.
+    */
+  private[config] def loadSuperuser(
       config: Config,
       permissions: Set[Permission]
   ): Either[String, ACL] = {
+    val path =
+      if (config.hasPath(superuserPath)) superuserPath
+      else if (config.hasPath(legacySuperuserPath)) legacySuperuserPath
+      else superuserPath
```
kelvin-chappell, line 133: Unless there's a point when you could have both, which shouldn't happen:

```suggestion
    val path =
      if (config.hasPath(legacySuperuserPath)) legacySuperuserPath
      else superuserPath
```

configTools/src/test/scala/com/gu/janus/config/LoaderTest.scala
```diff
@@ -172,19 +186,58 @@ class LoaderTest
...
+        .map(_.id) shouldEqual Set(
+        "website-s3-manager"
+      )
+    }
+
+    "prefers the superuser key when both keys are present" in {
+      val config = superuserBlock.withFallback(testConfig)
+      val userAccess = superuserAclFor(config).userAccess
+      userAccess.keySet shouldEqual Set("employee2")
+    }
```
kelvin-chappell, line 230: Is there a scenario where both keys will be present?

configTools/src/main/scala/com/gu/janus/config/Loader.scala
adamnfish-gu: Yeah I agree in practice, but I have to say Copilot CLI wasn't wrong when it recommended this. According to the spec either are supported so it's not wrong to handle both explicitly. One gotcha is that with your simplification (which I also looked at first), we'll be prioritising the legacy path if it exists. Again I don't mean to say this is necessarily wrong, but I think the idea is supposed to be "superuser is the correct config path, we fall back to the legacy admin path if needed". 

Worth saying this code is temporary - it's already scheduled for removal here:
https://github.com/guardian/janus-app/pull/935/changes#diff-38a21afeadc45b44cc46574e7dbdee0d37bdca041f87d5e0ffb4458debb3d6cfL130

So I don't mind it! I think it's terse and temporary enough that the extra accuracy isn't hurt by the slight extra verbosity.

What do you think?

configTools/src/test/scala/com/gu/janus/config/LoaderTest.scala
adamnfish-gu: As above. Practically no (we control the creation), but by the spec, yes. It could happen!

One plausible way is that a DevX/Security engineer is editing their local config, another is that some other company are using janus-app still and generate things differently to us. These are super edge cases that we shouldn't be worried about, but I do think the code as written here accurately reflects the spec. The simplification doesn't meet it quite as well. Especially since it's temporary, I'm ok to leave it. No harm changing it if you prefer!

configTools/src/main/scala/com/gu/janus/config/Loader.scala
kelvin-chappell: Yes agreed.  It's also temporary as you say

kelvin-chappell (approved):
👍

PR #934, opened 2026-07-29
Update the example project for janus-config-tools 13.0.0

adamnfish-gu:
> [!NOTE]
> **Disclaimer**: AI was used to significantly generate all the stacked PRs associated with this change. I've manually reviewed all the work and taken time to edit and curate the AI's output (both code and text), but this work has primarily been done by Opus 5 via the GitHub Copilot CLI.

Updates the example project for `janus-config-tools` 13.0.0.

- Bumps the pinned dependency from 11.0.0 to 13.0.0
- `Data.scala` uses `superuser = Superuser.acl`
- `DataTest` round-trips `janusData.superuser`

> [!WARNING]
> **CI on this PR will fail, and that is expected.** The example project resolves the published artifact rather than the in-repo module, so it cannot compile against the renamed API until 13.0.0 is on Maven Central. The same applies locally, so the diff is the only thing to review for now.
>
> **Do not merge until 13.0.0 is published.** That means #933 merged and the Release workflow run. Once the artifact is available, re-run CI here and it should go green.

---

Part of #937, which tracks the whole migration including the releases,
deploys and verification steps between these PRs.

1. app: rename estate-wide concept to superuser (#930)
1. app: rename `AccessSource.Admin` to `Superuser` (#931)
1. configTools: read both `superuser` and legacy `admin` keys (#932)
1. guardian/janus and other consumers: bump to 12.0.0
1. configTools: write `superuser`, rename `JanusData.admin` (#933)
1. example project pinned to 13.0.0 (#934) — CI red until 13.0.0 ships  ← **you are here**
1. guardian/janus and other consumers: bump to 13.0.0
1. configTools: remove legacy `admin` read support (#935)
1. guardian/janus and other consumers: bump to 14.0.0

PR #933, opened 2026-07-29
configTools: write the superuser key and rename JanusData.admin

adamnfish-gu:
> [!NOTE]
> **Disclaimer**: AI was used to significantly generate all the stacked PRs associated with this change. I've manually reviewed all the work and taken time to edit and curate the AI's output (both code and text), but this work has primarily been done by Opus 5 via the GitHub Copilot CLI.

Switches the writer to the new key and renames the field.

- `JanusData.admin` becomes `JanusData.superuser`
- The Twirl template emits a `superuser { ... }` block
- App usages and `configTools` tests are updated
- `example.conf` uses the new key; the legacy-key tests from #932 now inject the old key programmatically

> [!IMPORTANT]
> **Breaking change.** This changes both the `configTools` API and the generated config file format, and needs a major release (13.0.0) before the config repo can adopt it.

Reading is unaffected: #932 is already merged by this point, so the loader accepts both keys.

Note: This repo's `example/` project stays pinned to 11.0.0 here. It will be updated in a subsequent PR, once 13.0.0 is published.

---

Part of #937, which tracks the whole migration including the releases,
deploys and verification steps between these PRs.

1. app: rename estate-wide concept to superuser (#930)
1. app: rename `AccessSource.Admin` to `Superuser` (#931)
1. configTools: read both `superuser` and legacy `admin` keys (#932)
1. guardian/janus and other consumers: bump to 12.0.0
1. configTools: write `superuser`, rename `JanusData.admin` (#933)  ← **you are here**
1. example project pinned to 13.0.0 (#934) — CI red until 13.0.0 ships
1. guardian/janus and other consumers: bump to 13.0.0
1. configTools: remove legacy `admin` read support (#935)
1. guardian/janus and other consumers: bump to 14.0.0

PR #935, opened 2026-07-29
configTools: remove legacy admin key read support

adamnfish-gu:
> [!NOTE]
> **Disclaimer**: AI was used to significantly generate all the stacked PRs associated with this change. I've manually reviewed all the work and taken time to edit and curate the AI's output (both code and text), but this work has primarily been done by Opus 5 via the GitHub Copilot CLI.

Removes the legacy `janus.admin` fallback, so the loader accepts only `janus.superuser`.

- The fallback is dropped from `Loader.loadSuperuser`
- The legacy-key tests are removed
- The remaining test resources move to the new key

> [!IMPORTANT]
> **Breaking change.** This removes support for the old config file format and needs a major release (14.0.0).

> [!WARNING]
> **Blocked on sequencing.** CI here is green and the change is fully tested locally, but it must not merge until:
> - the config repo is generating the new format in production, and it has been stable for a while
> - developers have refreshed their local `janus.local.conf` with the `superuser` key
>
> After this, an old local config will fail to load.

---

Part of #937, which tracks the whole migration including the releases,
deploys and verification steps between these PRs.

1. app: rename estate-wide concept to superuser (#930)
1. app: rename `AccessSource.Admin` to `Superuser` (#931)
1. configTools: read both `superuser` and legacy `admin` keys (#932)
1. guardian/janus and other consumers: bump to 12.0.0
1. configTools: write `superuser`, rename `JanusData.admin` (#933)
1. example project pinned to 13.0.0 (#934) — CI red until 13.0.0 ships
1. guardian/janus and other consumers: bump to 13.0.0
1. configTools: remove legacy `admin` read support (#935)  ← **you are here**
1. guardian/janus and other consumers: bump to 14.0.0

### DIFF
--- a/configTools/src/main/scala/com/gu/janus/config/Loader.scala
+++ b/configTools/src/main/scala/com/gu/janus/config/Loader.scala
@@ -14,7 +14,6 @@ import scala.util.Try
   */
 object Loader {
   private val superuserPath = "janus.superuser"
-  private val legacySuperuserPath = "janus.admin"
 
   def fromConfig(config: Config): Either[String, JanusData] = {
     for {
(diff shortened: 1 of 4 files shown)


kelvin-chappell (approved):
👍

PR #938, opened 2026-07-30
Prevent the dependency-review workflow from blocking PRs

adamnfish-gu:
We've been seeing PRs that do not touch dependencies be blocked by security alerts. Let's make sure this status check isn't required, while keeping the information it provides to both the PR and dependency graph.


kelvin-chappell (approved):
👍

PR #942, opened 2026-08-03
Add CSV export option to the account audit trail

adamnfish-gu:
This allows teams to review historic access to see who still needs access.

Just a conversation starter, let's talk to the EMs as well. In particular, is 6 months right, would three months be better? Does this need to be configurable? I've assumed we don't need this for individuals, is that right?

Lots of questions, but this PR was easy to create as a starting point.

### How it looks
Screenshot 2026-08-03 at 16 54 07

### How the export gets imported
Screenshot 2026-08-03 at 16 56 54

### DIFF
--- a/app/aws/AuditTrailDB.scala
+++ b/app/aws/AuditTrailDB.scala
@@ -90,8 +90,8 @@ object AuditTrailDB {
       dynamoDB: DynamoDbClient,
       request: QueryRequest
   ): Seq[Either[String, AuditLog]] = {
-    val result = dynamoDB.query(request)
-    result
+    dynamoDB
+      .queryPaginator(request)
       .items()
       .asScala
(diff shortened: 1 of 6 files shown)


johnduffell:
great idea - I think if 6 months is going to be performant/cheap enough then why give less.  Since it's CSV that gives us the ability to drop in a sheet and do what we want with it.

What I actually need is "for each user with membership access at each level, when did they last use that access".

So the only thing missing would be people with access that they haven't used in the time period at all.  (I tried to use copilot to get that list from Access.scala and it did a terrible job)


What you've done would have been helpful for me although the time saving would be in the order of minutes.

adamnfish-gu:
Great feedback, it sounds like this isn't "the tricky part". I've started talking to the EMs about what would be useful to people that need to maintain our access configuration, and will be sure to reach out to others as well. Thanks again @johnduffell 🙏

PR #943, opened 2026-08-06
Force a major version bump in the janus-config-tools library

adamnfish-gu:
The automation in the previous release incorrectly issued that change as a "patch". It should have been a major version change.

This PR forces a major version bump of janus-config-tools by manually setting the snapshot to the next major version.

Thanks @rtyley for the pointer.


rtyley (approved):
Yep, this will force a major version bump!

